PR #29, opened 2026-07-01
feat: device grouping for batch operations (#5)

### LINKED ISSUE #5: Device grouping for batch operations (many-to-many)
Summary
=======
Add the ability to organize devices into named groups and target a whole group with
a single selection. A device can belong to multiple groups at the same time
(many-to-many). Selecting a group selects all its member devices for any existing
command (launch, install, startup app, push files, …).

Motivation
==========
During development we repeatedly operate on the same fixed set of ~10 headsets.
Re-selecting them by hand for every command is tedious and error-prone. Grouping
lets us define that set once and target it in a single click.

Requirements
============
- Create / rename / delete named groups.
- Assign a device to zero or more groups; one device can be in multiple groups
  simultaneously.
- Group membership is persisted server-side (survives restarts) and keyed by device
  serial, so a device can be assigned to a group even while offline
  (depends on PsychicVRLab/STYLY-MDM-PicoPsychicVRLab/STYLY-MDM-Pico#27).
- Selecting a group adds all its members to the current device selection; commands
  then dispatch to the online members only (reusing `resolve_target_ids()` /
  `target_devices`).
- Group membership is visible in the UI (which groups a device belongs to; which
  devices are in a group).

Suggested approach (reuse existing infra)
=========================================
- Store groups in the server-side registry (extend `device_registry.json`):
  e.g. `groups: { "<group name>": ["serialA", "serialB", ...] }`, persisted via the
  same atomic save pattern as labels.
- Add admin WebSocket messages mirroring `SET_DEVICE_LABEL`: e.g. `CREATE_GROUP`,
  `DELETE_GROUP`, `SET_DEVICE_GROUPS`.
- Include group info in the `DEVICE_LIST` payload (per-device `groups: [...]`)
  and/or a separate `GROUP_LIST` broadcast.
- Frontend: a group management section (collapsible, consistent with the existing
  command sections) plus a group selector that populates the existing `selectedIds`
  set. The command dispatch path is unchanged since it already accepts
  `target_devices`.

Considerations
==============
- Group selection can be a pure client-side convenience that expands to
  `target_devices` (smallest change), or the server could resolve
  `target_devices: ["group:dev-room"]` directly — decide which.
- Best built on top of the persistent device list (PsychicVRLab/STYLY-MDM-Pico#27) so offline members
  can be grouped and selected.
- Deleting a group must not delete devices.

hacha:
## Summary

Implements device grouping (many-to-many) for batch operations from #5, and
reorganizes the web console around the new grouping model (per the designer spec).

### Grouping (server + console)
- Named, many-to-many device groups persisted server-side in `device_registry.json`
  (wrapped `{devices, groups}` format; load stays backward-compatible with the legacy
  flat shape). Membership is keyed by **serial**, so offline devices can be grouped.
- New admin WebSocket messages: `CREATE_GROUP`, `RENAME_GROUP`, `DELETE_GROUP`,
  `SET_DEVICE_GROUPS`, `SET_GROUP_MEMBERS`; the server broadcasts `GROUP_LIST`.
  `FORGET_DEVICE` also removes the serial from all groups.
- Command dispatch (`resolve_target_ids` / `target_devices`) is **unchanged** —
  selecting a group sets the client selection to its (online) members.

### Console redesign
- Three-zone layout: **Targets** (Groups | Devices sub-tabs, shared selection),
  **Actions** (selection summary + Quick Install / Launch / Startup), **Activity Log**.
- Groups are single-select and drive the selection; a group-centric **Manage Groups**
  screen handles create / rename / delete / membership.
- Full-width console with **draggable** Targets|Actions and zones|Log boundaries
  (sizes persisted in `localStorage`).
- Actions commands collapse into an **accordion** (one open at a time, all closed by default).
- **Group rows expand** to show each member's online/offline status and Quick Install progress.
- English-only UI strings; the no-groups state is usable (defaults to the Devices tab,
  always-available "Manage Groups" entry).

Note: the redesign supersedes the "remove redundant to All buttons" change (#28,
already on `develop`); the rewritten console has no such buttons.

## Test plan
- **Server logic**: registry load/save (both formats), all group handlers + validation,
  forget cleanup, `SET_GROUP_MEMBERS` (offline-allowed, dedupe, error cases).
- **Frontend (jsdom on the real `index.html`, 53 checks)**: sub-tabs & default tab,
  group single-select, device table, label/forget, launch/install dispatch,
  Manage Groups CRUD, group expand + live install status, Actions accordion.
- **End-to-end** (live server + fake device): membership persisted; `EXECUTE_LAUNCH` /
  `EXECUTE_INSTALL` dispatched to online members.
- **Visual** (real Chrome via Playwright): desktop & narrow layouts, populated states,
  drag cursor-tracking.

Closes #5

### DIFF
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -149,6 +149,11 @@ STYLY-MDM/
 | `LAUNCH_APP` | Launch an app on target devices. Fields: `target_devices` (list of device IDs or `["*"]`), `package_name`, `extra_data` |
 | `INSTALL_APK` | Install an uploaded APK on target devices. Fields: `target_devices` (list of device IDs or `["*"]`), `apk_url`, `apk_filename` |
 | `GET_DEVICE_LIST` | Request the current device list |
+| `CREATE_GROUP` | Create a new, empty device group. Fields: `name` |
+| `RENAME_GROUP` | Rename a group, preserving its members. Fields: `name`, `new_name` |
+| `DELETE_GROUP` | Delete a group (member devices are not affected). Fields: `name` |
+| `SET_DEVICE_GROUPS` | Set the exact set of groups a device belongs to. Fields: `device_id`, `groups` (list of existing group names) |
+| `SET_GROUP_MEMBERS` | Set the exact member list of an existing group (group-centric). Fields: `name`, `members` (list of serials; offline/unknown serials allowed) |
 
 ### Admin HTTP API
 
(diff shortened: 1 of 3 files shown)


hacha:
P1: active group selection is not recomputed after `GROUP_LIST` updates, so commands can still be sent to a stale selection after editing the selected group.

Reproduction:
1. Select group A in the Groups tab.
2. Open Manage Groups, change group A's members, and save.
3. Return to the console.

At this point the UI still says `via group "A"`, but `selectedIds` still contains the members from when the group was originally selected. As a result, Launch / Install / Startup can be sent to the old targets instead of the current members of group A.

When `GROUP_LIST` is received, if `activeGroup` still exists, `selectedIds` should be rebuilt from the current `groups[activeGroup]`. If the active group no longer exists, clearing both `activeGroup` and the selection seems like the safest behavior.

hacha:
Fixed in 7c4184f.

`GROUP_LIST` and `DEVICE_LIST` now recompute the active-group selection via a new `syncActiveGroupSelection()`:
- if `activeGroup` still exists, `selectedIds` is rebuilt from the current `groups[activeGroup]` (intersected with known devices), so editing the selected group's members immediately updates the targets;
- if the active group no longer exists, both `activeGroup` and `selectedIds` are cleared.

Added jsdom coverage: selecting a group, changing its membership (selection recomputed), and deleting the active group (selection cleared).

PR #48, opened 2026-07-11
feat: self-update the client over MDM with power-cycle recovery (#39)

### LINKED ISSUE #39: feat: self-update the HMD client over MDM (survive package replacement and auto-restart)
## Goal

Allow the MDM client on the HMD to be **updated over MDM** instead of by cable. Initial provisioning by cable/sideload is acceptable; every subsequent update should be pushed from the management server.

Scope of this issue is the client **updating itself** (`com.styly.mdmclient`). Updating a *separate* VR content app is a much simpler case — the caller process survives, `IIntCallback` fires, and `pbsStartActivity` can relaunch it — and is not covered here.

## Current behavior

- Silent install already works: `MdmClientService.installApk` (`mdm-client/app/src/main/java/.../MdmClientService.kt:278-313`) calls
  `ToBServiceHelper.getInstance().serviceBinder.pbsControlAPPManger(PBS_PackageControlEnum.PACKAGE_SILENCE_INSTALL, path, 0, IIntCallback)`.
- When the target package is the client itself, Android kills the client process on package replacement. This is unavoidable.
- Consequences of the self-kill:
  - The `IIntCallback` never fires on success (the binder dies with the process), so the current success path reports nothing.
  - The post-install cleanup of the downloaded APK under `Downloads/styly-mdm/` never runs.
  - The WebSocket drops, and the server currently interprets the disconnect as "offline" rather than "updating".
- There is **no** `ACTION_MY_PACKAGE_REPLACED` receiver. `BootReceiver` (`AndroidManifest.xml:55-62`) only handles `BOOT_COMPLETED` and `com.pvr.tobservice.SERVICE_AUTO_BOOT`.
- `downloadApk` (`MdmClientService.kt:213-276`) fetches the APK over plain LAN HTTP with **no integrity check** before handing the path to the silent installer.

## Spike first: two questions to answer on a real device

The design below depends on device behavior we have not measured. Answer these before writing the implementation.

1. **Does `pbsAppKeepAlive` already relaunch the client after a self-update?**
   The app registers PICO keep-alive at `MdmClientApplication.kt:29`. If keep-alive already covers package replacement, recovery may need no new code.

2. **Does `startForegroundService()` from a `MY_PACKAGE_REPLACED` receiver succeed?**
   Android 12+ restricts starting a foreground service from the background. `BOOT_COMPLETED` is a documented exemption; whether `ACTION_MY_PACKAGE_REPLACED` is exempt needs verification, and PICO devices span Android 10-14. If it throws `ForegroundServiceStartNotAllowedException`, fall back to the ToBService keep-alive / `pbsStartActivity` path instead of a direct FGS launch.

### Restart

Add `android.intent.action.MY_PACKAGE_REPLACED` to a manifest-declared receiver. This broadcast is delivered to the *new* APK right after its own package is replaced, and manifest receivers are permitted for it.

### Result reporting: persist-then-confirm, not callback

Since the callback cannot be relied on for success:

1. Before invoking silent install, persist `{ update_in_progress, target_version_code, correlation_id }` to `stylymdm_prefs`.
2. Invoke silent install. The process dies.
3. The new APK receives `MY_PACKAGE_REPLACED`, starts the service, reconnects the WebSocket.
4. Read the persisted state, compare against `BuildConfig.VERSION_CODE`, and report the outcome to the server with the correlation id.
5. Clean up the leftover APK in `Downloads/styly-mdm/`.

Both paths must be handled: on **failure** the process does not die and the `IIntCallback` *does* fire with a PICO result code. A timeout is also needed for the "install invoked but nothing happened" case.

Server side: treat the disconnect during a known update as `updating`, not `offline`, and wait for re-registration. Include `versionCode` in the `REGISTER` payload so the server can confirm the new version.

### Safety (the load-bearing part)

A bad update bricks the client: if the new APK crashes on launch we lose remote control and the device needs cable recovery. Android offers no practical rollback (downgrade requires uninstall, which wipes app data). Therefore:

- **Verify before installing.** The server sends the expected `cd_sha256` / `full_sha256` / `signer_sha256` with the install command; the client refuses to invoke the installer on mismatch. This reuses the hashing already implemented for #37 and closes the plain-HTTP gap in `downloadApk`. This is the highest-value single change in this issue.
- **Staged rollout.** Push to one device and confirm before the fleet.
- **Verify after restart.** Run `EXECUTE_VERIFY_APK` against the client's own package to confirm the installed APK matches the target, closing the loop.

## Known pitfalls for testing

Silent install failures surface as PICO-specific result codes via `installResultMessage` (`MdmClientService.kt:427-442`), not readable Android errors. The two we will almost certainly hit:

- **Signature mismatch.** A locally built debug-signed APK cannot replace a CI release-signed install. Worse, the `prod` and `dev` product flavors share `applicationId com.styly.mdmclient` (`app/build.gradle:7`), so they replace each other.
- **`versionCode` must strictly increase.** No downgrade.

## Acceptance criteria

- [ ] Spike results for the two questions above are recorded in this issue.
- [ ] The client refuses to install an APK whose hash does not match the reference supplied by the server.
- [ ] After an MDM-triggered self-update, the client restarts unattended and reconnects to the server.
- [ ] The server observes `updating` (not `offline`) across the restart and receives an update-result report carrying the correlation id and the new `versionCode`.
- [ ] Post-update self-verification confirms the installed APK matches the target.
- [ ] The stale APK under `Downloads/styly-mdm/` is removed after the restart.
- [ ] `docs/DEVELOPMENT.md` is updated with the self-update flow (and `message-flow.drawio.svg` if the message sequence changes).

## Related

- #37 — integrity verification (hashing reused here)

hacha:
## Summary

Closes #39. Lets the MDM client update **itself** over the same push-install path used for
other apps, and recover automatically afterwards.

The spike that opened this branch established, on a real PICO, that **nothing restarts the
client** after the package installer kills its process: `pbsAppKeepAlive` does not survive
package replacement and `MY_PACKAGE_REPLACED` is never delivered. So recovery cannot rely on
an in-process hook — the dying build **arms a one-shot device power cycle before invoking the
installer**, and the replacement build disarms it after it boots.

## How it works

1. **Verify before betting the process.** The download travels over plain LAN HTTP, so the
   server ships the APK's SHA-256 with `EXECUTE_INSTALL` and the client refuses on mismatch.
   A *self*-update additionally refuses when the hash is **absent** (old server) — a bad
   client build costs remote control and Android has no rollback, so it never installs
   unverified bytes over itself.
2. **Arm the power cycle.** `openTimingShutdown` at +2 min, `openTimingStartup` at +3 min
   (`PowerCycleSchedule`, 1-based month, host-JVM tested). Scheduling failure refuses the
   update rather than stranding the device.
3. **Announce, then install.** Persist the update marker, send `SELF_UPDATE_STARTING`, flush
   the outbound queue, invoke the silent installer; the process dies.
4. **Server renders `updating`.** The disconnect after `SELF_UPDATE_STARTING` reads as
   `updating`, not `offline`. Bounded by `MDM_SELF_UPDATE_TIMEOUT` (480 s).
5. **Reboot recovery.** The new build boots, disarms the timers, sweeps the leftover APK,
   reconnects, and re-registers with its `version_code`.
6. **Result + auto-verify.** The server settles the update (`SELF_UPDATE_RESULT`), then runs
   `EXECUTE_VERIFY_APK` against the client's own package and compares the hash
   (`SELF_UPDATE_VERIFIED`). That `VERIFY_APK_RESULT` is consumed server-side, never
   forwarded. Bounded by `MDM_SELF_UPDATE_VERIFY_TIMEOUT` (120 s).

On **install failure** the process survives and the callback fires: the client disarms the
timers (unconditionally — the callback firing proves the process was not replaced) and sends
`INSTALL_RESULT` fail, which clears the server's pending state.

## What's in the diff

- **Client:** `InstallPolicy` (hash gate), `PowerCycleSchedule` / `PowerCycleTimers`, recovery
  in `MdmClientApplication`, self-update flow in `MdmClientService`, `version_code`/
  `version_name` in REGISTER. Removes the spike scaffolding (`PackageReplacedReceiver`,
  `spikeMode`, `SPIKE_MODE`); the journal, in-headset viewer, and start-reason plumbing stay
  as permanent diagnostics.
- **Server:** reference-hash dispatch (stat-keyed cache), `pending_self_updates` lifecycle,
  auto-verify, two timeouts.
- **Console:** `updating` row/cell, command targets restricted to strictly-online devices,
  client version per row, `SELF_UPDATE_*` log lines.
- **Docs:** `DEVELOPMENT.md` self-update flow, README bullet, `message-flow.drawio.svg` lane.

## Testing

- **Server:** 181 pass, incl. `tests/test_self_update.py` (e2e happy path, double-command,
  disconnect + slot release, timeout, verify timeout).
- **Client:** host-JVM `InstallPolicyTest` / `PowerCycleScheduleTest`; `compileDevDebugKotlin`
  builds.
- **Device:** runs A–E on PICO A9210 (PUI 5.15.5 / Android 14, API 34) — spike questions
  answered, and the full happy-path self-update + power-cycle recovery verified end to end.


hacha:
I found one blocking issue and one follow-up issue.

**P1 blocking: timer cleanup can retire the retry flag even when TobService refuses to close a timer.** `PowerCycleTimers.disarm()` only treats thrown exceptions as cleanup failure, then clears the persisted `power_cycle_timers_set` flag whenever the close calls do not throw. However, `closeTimingShutdown()` and `closeTimingStartup()` are `int` return-value APIs in `tobservicelib-4.3.16`, just like the open calls. If `openTimingShutdown()` succeeds, `openTimingStartup()` fails, and the cleanup `closeTimingShutdown()` returns non-zero without throwing, the self-update is refused but the shutdown timer can remain armed with no persisted retry flag. That can strand the device in the exact failure path this guard is trying to avoid.

Please treat non-zero close returns as failed cleanup and keep the retry flag set until both close calls return success.

**P2 follow-up: post-update auto-verify should use the dispatched job's hashes, not a later filename echo.** `EXECUTE_INSTALL` hashes the uploaded APK before dispatch, but `SELF_UPDATE_STARTING` recomputes `expected_full_sha256` from the client-supplied `apk_filename`. If that echo is empty/wrong, or a same-name APK is replaced between dispatch and announcement, verification can be skipped or compared against the wrong local file. It would be more robust to preserve the hashes from the install dispatch per target/job, or validate the echo against active install state before using it as the verify reference.

Verification I ran:

- `python -m pytest -q` from `mdm-server/`: 181 passed
- `python -m pytest -q tests/test_self_update.py` from `mdm-server/`: 22 passed
- `JAVA_HOME=/opt/homebrew/opt/openjdk ./gradlew :app:testDevDebugUnitTest :app:compileDevDebugKotlin --no-daemon` from `mdm-client/`: passed
- `git diff --check origin/develop...HEAD`: passed

I also checked the current PICO Business docs via Context7 for TobService/high-risk API context; exact timer signatures were sparse there, so I verified the cached `tobservicelib-4.3.16` AAR with `javap`.

hacha:
Addressed both findings in 4f3aa6c.

**P1 (blocking) — timer disarm now treats non-zero close returns as failed cleanup.**
`closeTimingShutdown()`/`closeTimingStartup()` are `int`-returning (confirmed via `javap` on the `tobservicelib-4.3.16` AAR). `PowerCycleTimers.disarm()` now clears the persisted armed flag only when both closes neither throw nor return non-zero; any unconfirmed close keeps the flag set so the next process start retries.

One deliberate deviation from "keep the retry flag set until both close calls return success": I added a `MAX_DISARM_ATTEMPTS = 3` cap. Retrying a *fired* one-shot timer forever would loop the recovery on every boot (a past-due timer can no longer fire), so after 3 starts the flag is retired anyway. `POWER_CYCLE_CLOSED` now records `cleared` / `retry_pending attempts=N` / `gave_up attempts=N` so the outcome is diagnosable from the headset.

**P2 (follow-up) — verify reference pinned to the dispatched bytes.**
The server now records the hashes it dispatched to each device (`last_install_dispatch`) at `EXECUTE_INSTALL` time and verifies the self-update against those, instead of re-hashing the client-echoed `apk_filename`. A same-name re-upload — or a wrong/empty echo — between dispatch and announcement can no longer shift the reference. Falls back to re-hashing only when there is no dispatch record (e.g. server restarted mid-update); the record is evicted on disconnect. New e2e test asserts the pinned reference survives a same-name re-upload.

**Two things to flag honestly:**
1. **P1 is compile + logic-reviewed, not device-verified.** E1 was verified with the *old* disarm, and this changes that exact recovery path (the journal line on the happy path may now read `retry_pending`/`gave_up` instead of `cleared` depending on what a fired timer's close returns). Please re-run E1 to confirm — ideally `cleared`; `retry_pending → gave_up` is the acceptable fallback.
2. **This does not close the `arm()` partial-failure strand.** If `openTimingShutdown` succeeds, `openTimingStartup` fails, and the cleanup `closeTimingShutdown` also fails, the device powers off at +2 min with no startup timer and never reboots, so the flag-retry (which only runs at process start) never fires. Only an in-process close-retry inside `arm()` before refusing would close that gap — left out of scope here; happy to open a follow-up if you want it.

**Verification:**
- `python -m pytest -q` from `mdm-server/`: **182 passed** (+1)
- client `:app:testDevDebugUnitTest :app:compileDevDebugKotlin`: passed
- `git diff --check origin/develop...HEAD`: passed
- Docs: `DEVELOPMENT.md` self-update steps 5 (disarm confirm/retry) and 6 (pinned verify reference) updated.

hacha:
Thanks for pushing `4f3aa6c`. I re-checked the PR head and confirmed it now includes the fix commit.

The P2 auto-verify issue looks resolved: `last_install_dispatch` now pins verification to the hashes captured at `EXECUTE_INSTALL` dispatch time, and the same-name re-upload test covers the important regression case.

One blocking risk remains from the original P1, though: the `arm()` partial-failure path is still not closed. If `openTimingShutdown()` succeeds, `openTimingStartup()` fails, and the cleanup `closeTimingShutdown()` also cannot confirm success, the client refuses the self-update but can leave a shutdown timer armed with no startup timer. Because the device may power off before any reboot/recovery path exists, the retry-on-next-process-start logic in `disarm()` cannot help in that specific case.

Please handle that failure path inside `arm()` before returning `false`. For example, after a partial schedule failure, retry closing any timer that may have been opened until the close is confirmed or a short bounded retry window expires; if cleanup still cannot be confirmed, the code should surface that as an unsafe state rather than treating the refused install as enough recovery. The key invariant is: `arm()` must not return `false` while a shutdown timer may still be armed without a paired startup timer.

Verification after the pushed commit:

- `python -m pytest -q` from `mdm-server/`: 182 passed
- `JAVA_HOME=/opt/homebrew/opt/openjdk ./gradlew :app:testDevDebugUnitTest :app:compileDevDebugKotlin --no-daemon` from `mdm-client/`: passed
- `git diff --check origin/develop...HEAD`: passed

hacha:
Handled the `arm()` partial-failure path in 694263a.

**`arm()` now closes out a partial failure in-process instead of relying on the boot-time retry.** You're right that the flag-retry can't help here — it only runs at the next process start, which a device stranded by a lone shutdown timer never reaches. So:

- `arm()` now returns an `ArmOutcome` (`ARMED` / `REFUSED_SAFE` / `REFUSED_UNSAFE`) instead of a bare `Boolean`.
- On any scheduling failure it closes **both** timers in-process (the open return codes aren't trusted to say what actually armed) with a bounded retry — `MAX_ARM_CLOSE_ATTEMPTS = 4`, 150 ms backoff — so a transient close failure is absorbed. `arm()` runs off the main thread (inside `installApk`'s `Thread`), so the short blocking backoff is ANR-safe.
- The refusal is reported `REFUSED_SAFE` only once the shutdown timer is confirmed closed; otherwise `REFUSED_UNSAFE`, and `MdmClientService` sends an `INSTALL_RESULT` fail whose detail flags the unsafe state ("...a shutdown timer may still be armed without a paired startup timer; the device may power off until it is manually restarted"), so a still-reachable device surfaces it to the console rather than treating the refused install as enough recovery.
- The invariant is enforced against the *actual* danger only: the verdict is gated on whether shutdown was really armed (`open == 0`) and on a sticky per-timer "ever confirmed closed", so (a) a close of a never-armed shutdown reporting non-zero can't raise a false unsafe alarm, and (b) a re-close of an already-closed one-shot can't flip a safe state back to unsafe.

The decision/retry logic is extracted as a pure `resolvePartialArm` and unit-tested on the host JVM (`PowerCycleResolveTest`, 7 cases: both-closed-early-stop, armed-shutdown-never-closes → unsafe, transient-recovery → safe, lone-lingering-startup → safe, sticky-reclose → safe, never-armed-shutdown → safe).

**Two honest caveats:**
1. **Only the verdict logic is tested.** The live wiring — the real proxy close calls, the backoff, the flag write, and the `MdmClientService` branch that sends the UNSAFE `INSTALL_RESULT` — has no test, and the partial-arm path is impractical to device-verify without fault injection (you'd have to force `openStartup` to fail while `openShutdown` succeeds). Same not-device-verified caveat as P1 still applies.
2. **I did not add active startup-pairing as a fallback.** Retrying the `openTimingStartup` that just failed buys little, so I stuck to your "retry-close, else surface as unsafe" ask. No client code can force an unresponsive timing API to close — a persistent failure is *surfaced to the console*, not eliminated. If you'd rather `arm()` also attempt to arm a paired startup as a last resort (an unwanted-but-recoverable power cycle vs. an unpaired shutdown), say so and I'll add it.

**Verification (this commit):**
- client `:app:testDevDebugUnitTest :app:compileDevDebugKotlin`: passed (`PowerCycleResolveTest` 7/7)
- `git diff --check origin/develop...HEAD`: passed
- Docs: `DEVELOPMENT.md` self-update step 2 (partial-failure close-out + unsafe surfacing) updated.

PR #51, opened 2026-07-11
fix: recover client self-updates with an embedded guard app (#39 follow-up)

hacha:
## Problem

The self-update shipped in #48 never worked on real hardware — silently:

1. **Missing Gson dependency.** The PICO timing APIs (`openTimingShutdown`/`openTimingStartup`) resolve `com.google.gson.JsonObject` internally, but `tobservicelib`'s AAR ships no POM, so Gson was never pulled transitively. Arming the power-cycle recovery raised `NoClassDefFoundError` — an `Error`, invisible to every `catch (Exception)` — which killed the install thread before `SELF_UPDATE_STARTING` was sent or anything was journalled. No journal entry, no server signal, no install. On-device crash records (`logcat -b crash`, dropbox) showed three identical crashes across independent attempts.
2. **The power-cycle recovery is dead on current firmware.** With Gson fixed, the timing APIs return `-1`: the firmware's SELinux policy denies the poweroffalarm app access to the Qualcomm RTC alarm HAL (`avc: denied { find }` for `vendor.qti.hardware.alarm.IAlarm`, `permissive=0`). Not fixable app-side, so a self-update had no working way back and was always refused.

## What this PR does

- **Fixes the crash** (`acc9f96`): adds the Gson dependency and widens SDK-boundary guards to `catch (Throwable)`, so a pre-announcement failure surfaces as a reported `INSTALL_RESULT` fail instead of a silent process death.
- **Replaces the recovery mechanism with a guard app** (`d8fc14b`): `guard/` (`com.styly.mdmguard`) is a separate package the client's self-replace cannot kill. It binds TobService, checks the client's process every 10 s, and revives a dead client through the privileged `startForegroundService` — exempt from background-start restrictions, punches through the stopped state. The self-update gate prefers the guard (confirming its process before the installer runs, journalled as `recovery=guard`) and falls back to the power-cycle timers where firmware allows them; only when both are unavailable is the update refused. Server and console are untouched — the existing `SELF_UPDATE_STARTING` / `updating` / `SELF_UPDATE_RESULT` / `SELF_UPDATE_VERIFIED` flow just completes in seconds instead of requiring a ~3 min device reboot.
- **Zero-touch guard deployment** (`ead6b07`): the guard APK ships embedded in the client (`assets/guard.apk`); the client's mutual-watch tick silent-installs it when missing or outdated and starts it when down. Deploying the client is the whole deployment story, a directly-deleted guard comes back within a minute, guard upgrades ride client self-updates, and `assembleProdRelease` now covers the guard (same env-driven signing as `:app`).
- **Lifecycle coupling** (`cc25e51`, `c6a1465`): the guard has no launcher entry, so an uninstalled client would otherwise leave an invisible orphan. When the client's package goes missing (a true uninstall — a replace never reads as missing), the guard stands down and self-uninstalls after 10 minutes of accumulated absence. The absence clock is persisted (survives process death and reboots) and a self-chaining deadman alarm (`setAndAllowWhileIdle` + WAKEUP, hourly) restarts the guard through process death and device sleep — field testing caught exactly that gap.
- **Staging hardening** (`54a0d99`, from security review): the embedded guard APK must be staged in shared storage (the installer runs as a separate system user and fails with 102 on app-private paths), so the staged name is randomized and the bytes are re-hashed against the asset's SHA-256 immediately before the privileged install.

## Verification (all on a real PICO 4 Ultra, DEV-001, over the real MDM path)

- Full self-update E2E, three times (v10→v11, v11→v12, v13→v14 including a guard upgrade riding along): Quick Install → `updating` → installer kills the process → guard revival in ~3 s (journal: `recovery=guard`, new build's `APP_ONCREATE`, `SELF_UPDATE_CONFIRMED`, `reason=guard`) → `SELF_UPDATE_RESULT success` → `SELF_UPDATE_VERIFIED`, ~4–7 s from dispatch to verified.
- Revival after `am force-stop` and after package replacement; guard auto-reinstall after direct deletion; guard upgrade 1→2 via embedded copy.
- Self-destruct: verified including crashing the guard mid-grace (`am crash`) — the restarted process resumed the persisted clock and removed the package exactly at the accumulated grace; deadman alarm registration confirmed via `dumpsys alarm`.
- Client host-JVM unit tests pass; server tests untouched.

## Deployment notes

- **Deploy the server before pushing a new client** (unchanged): a self-update refuses to run without server-supplied hashes.
- **The first rollout is manual**: the running build executes the update, so clients whose running build predates this branch cannot self-update onto it — one cable/attended install per device.
- `docs/DEVELOPMENT.md` and `docs/message-flow.drawio.svg` are updated in the same change.

## Follow-ups (tracked in tasks/todo.md, not in this PR)

- The pre-existing download-and-install path stages APKs with the same shared-storage exposure the guard staging was hardened against; apply the same randomize + last-moment re-hash there.
- Provisioning docs should call out that `MANAGE_EXTERNAL_STORAGE` is a per-UID appop (a full uninstall+reinstall of the client drops it).


hacha:
Blocking finding: successful guard-based self-updates leave the downloaded client APK behind in shared storage.

The new primary recovery path sets `recovery=guard` when `GuardLink.ensureRunning(...)` succeeds, but that path does not set the persisted power-cycle armed flag. After the replacement build starts, `MdmClientApplication.recoverFromSelfUpdate()` returns immediately unless `UpdateJournal.powerCycleTimersSet(this)` is true, so `cleanupDownloadedApks()` only runs for the power-cycle fallback. On a successful self-update the `IIntCallback` path cannot clean up either, because the client process is killed by package replacement before the callback is delivered.

That means the guard-based success path leaves the downloaded APK under `Downloads/styly-mdm/`, which violates issue #39's stale-APK cleanup acceptance criterion and contradicts `docs/DEVELOPMENT.md`, which says the replacement build deletes the leftover APK after restart. Please make post-update cleanup independent from the power-cycle armed flag, or otherwise persist enough recovery state for the guard path to sweep the downloaded APK after the new build starts.

hacha:
Good catch — confirmed real on the test device: `Downloads/styly-mdm/` had accumulated 24 leaked APKs (~206 MB) across this branch's own guard-path E2E runs and the earlier crash-era attempts.

Fixed in a369630. `confirmSelfUpdateIfLanded` now reports whether it just confirmed a landing, and `MdmClientApplication.onCreate` sweeps the download directory on that signal — the first run of a replacement build, independent of which recovery mechanism brought it back, and with no binder needed. The bind-callback sweep stays for the power-cycle fallback, where an armed flag with no landing means a failed install whose process died before the `IIntCallback` could delete its download; `docs/DEVELOPMENT.md` step 5 updated to match.

Verified end-to-end on DEV-001 over the real MDM path: self-update v17→v18 landed (`SELF_UPDATE_RESULT success`, `SELF_UPDATE_VERIFIED`), and the sweep removed all 25 APKs (the 24-file backlog plus the update's own download) — the directory is now empty.

Incidentally, the same session field-verified the rest of the lifecycle machinery: the client had been uninstalled on the device overnight, and the guard's persisted absence clock fired the self-destruct exactly at the 10-minute grace (`Client gone for 600s; uninstalling self`); the freshly side-loaded client then auto-provisioned a new guard from its embedded copy within one tick, and the self-update gate refused safely in the short window before that provisioning completed.

PR #52, opened 2026-07-11
feat: retire devices remotely via client self-uninstall (#49)

### LINKED ISSUE #49: Removing mdm-client before device handover requires physical access to every unit
## Problem

Venue deployments involve 100+ devices that are set up through STYLY-MDM (APK installs, content pushes), repeated over several content-adjustment rounds. At final handover the MDM client must not remain on the devices. An orphaned client keeps looking for a server that no longer exists — a WebSocket reconnect loop with exponential backoff capped at 30 s (`WebSocketManager.kt:45-46`), each attempt preceded by a UDP broadcast discovery packet to `255.255.255.255` (`ServerDiscovery.kt`) — indefinitely, on the customer's network. Even this light traffic is a concern for a delivered installation, and an unmanaged remote-control agent should not ship with delivered hardware in any case.

Today the only way to remove mdm-client is per-device physical work (`adb uninstall com.styly.mdmclient`) — at exactly the fleet size where MDM was adopted to avoid per-device handling.

## Current state

- The client installs apps via ToBService `pbsControlAPPManger(PACKAGE_SILENCE_INSTALL, …)` (`MdmClientService.kt:406`); nothing in the codebase uninstalls any package.
- The SDK version already shipped in the client (`tobservicelib:4.3.16`) also defines `PBS_PackageControlEnum.PACKAGE_SILENCE_UNINSTALL`.
- An uninstall is executed by ToBService — a PICO system service running in a separate process — not by mdm-client itself. The self-update spike (#39) showed ToBService accepts package operations that target the caller's own package: silent self-install succeeded in every run.
- mdm-client is not a device owner or device admin and holds no launcher role, so no teardown is required beyond the uninstall itself.
- Unlike self-update (#39), no recovery machinery is needed: after a successful self-uninstall, nothing is supposed to come back. The power-cycle timers stay untouched.

## Approaches

- **Self-uninstall command (preferred).** New console/server command (e.g. `RETIRE_DEVICE` → client `EXECUTE_SELF_UNINSTALL`) that calls `pbsControlAPPManger(PACKAGE_SILENCE_UNINSTALL, <own package>, …)`.
  - Fire-and-forget by construction: the process dies mid-uninstall, so no result callback can arrive. The client announces first (same pattern as `SELF_UPDATE_STARTING` in #39): send `SELF_UNINSTALL_STARTING`, then invoke the uninstall.
  - The server treats "announced, disconnected, never re-registered" as success; a device that re-registers after announcing is the failure signal.
  - The server should auto-forget the device (existing `FORGET_DEVICE` path) or park it in a terminal `retired` state so the device list does not accumulate permanently-offline entries.
  - The console needs a deliberately heavy confirmation UI: the operation is remotely irreversible — recovery means physical adb access per unit.
- **Status quo / fallback.** Manual `adb uninstall` per device.

## Open questions (device spike needed, same shape as #39)

1. Does ToBService actually allow `PACKAGE_SILENCE_UNINSTALL` of the calling package? Self-*install* is proven; self-*uninstall* is unverified.
2. Does the startup-app setting (`SET_STARTUP_APP`, PICO system state) survive removal of mdm-client? The handover flow depends on delivered content still auto-launching after the client is gone.
3. What residue is acceptable? Files pushed to shared storage remain by design; the `MANAGE_EXTERNAL_STORAGE` appops grant points at a package that no longer exists. Confirm nothing else lingers.

hacha:
Closes #49.

## Problem

At final venue handover the MDM client must not remain on delivered devices: an orphaned client reconnect-loops (30 s capped backoff) and UDP-broadcasts discovery packets on the customer's network indefinitely, and an unmanaged remote-control agent should not ship with delivered hardware. Removal used to mean per-device `adb uninstall` — at exactly the fleet size MDM was adopted to avoid.

## What this adds

A console-triggered **Retire Devices** command that makes selected clients silently uninstall themselves, using the same `pbsControlAPPManger(PACKAGE_SILENCE_UNINSTALL, <own package>)` primitive the guard's self-destruct already uses (device-verified there, including that ToBService accepts an uninstall of the calling package).

**The success signal inverts the self-update's: silence is success.** A removed client can send nothing, so the server declares success only when the device announced `SELF_UNINSTALL_STARTING`, disconnected, and stayed away for `MDM_RETIRE_TIMEOUT` (default 120 s). Both failure modes surface well inside that window: a client whose uninstall never ran still holds its WebSocket (checked directly at the deadline), and a merely-killed client is restarted by the ToBService keep-alive and re-registers within the ≤30 s backoff.

### Server (`server.py`)
- `RETIRE_DEVICE` admin command → per-device `EXECUTE_SELF_UNINSTALL` with a server-generated correlation id; `RETIRE_SENT` ack, terminal `RETIRE_RESULT`.
- `pending_retires` state machine: disconnect renders the row `retiring`; the timeout settles it. Deliberately in-memory — a server restart mid-window can never mark a device retired spuriously (it just shows `offline`).
- Success parks the registry record in a terminal, persisted `retired` state (`_coerce_record` now roundtrips the flag — it drops unknown keys, so the flag would otherwise vanish on restart). Retired rows are greyed, structurally excluded from command targets, and removable with the existing forget button; a manually reinstalled device that re-registers is re-adopted.
- A retire supersedes a pending self-update; forgetting a device cancels its pending retire.

### Client (`MdmClientService.kt`)
- `executeSelfUninstall`: pause guard provisioning → **uninstall the guard first** (bounded 15 s wait, proceed regardless — left to its own self-destruct the guard would linger 10 min–1 h and could fire one futile revive into the kill window) → best-effort keep-alive deregistration → announce + socket flush → self-uninstall.
- The uninstall callback can only ever report failure (success kills the process first); a failed retire clears the in-progress flag so the next mutual-watch tick reinstalls the guard — a failed retire never leaves the device unguarded.
- No recovery armed, nothing persisted: after a successful retire nothing is supposed to come back.

### Console (`index.html`)
- New bottom "Retire Devices" section: warning + "I understand" checkbox gating the button + a native `confirm()` listing the target devices.
- `retiring` / `retired` status rendering; `isOnline()` flipped from a `!== 'offline'` denylist to an allowlist so new statuses default to dimmed/not-targetable.

### Docs
- `DEVELOPMENT.md`: protocol tables + a new "Device Retirement" section; `README.md`: feature list + `MDM_RETIRE_TIMEOUT`; `message-flow.drawio.svg`: retire sequence (mxfile + rendered SVG).
- Documented a finding that corrects an assumption in #49: the console's **Startup App feature is client state** (SharedPreferences + launch from the client's own service), not PICO system state — it does **not** survive a retire. A handover that needs boot-time auto-launch must configure the device's boot app by other means (candidate follow-up issue).

## Verification

- `pytest` (mdm-server): **196 passed**, including the new `tests/test_retire.py` (10 unit + 4 E2E over real WebSockets: happy path, re-register-within-window failure, client-reported failure, retire-supersedes-self-update).
- Gradle: `:app:compileDevDebugKotlin`, `:guard:compileDebugKotlin`, `:app:testDevDebugUnitTest` all green.
- On-device (PICO, real fleet flow): retire executed from the console → client removed from the device → row went `Retiring…` → `RETIRE_RESULT success` after the 120 s window → row `retired`.


hacha:
P2: The retire failure path leaves the client keep-alive disabled.

The client disables TobService keep-alive before invoking its self-uninstall:

- mdm-client/app/src/main/java/com/styly/mdmclient/MdmClientService.kt:743-747

If the uninstall fails, failRetire() reports the failure and clears retireInProgress, but it does not re-enable keep-alive:

- mdm-client/app/src/main/java/com/styly/mdmclient/MdmClientService.kt:798-806

Since these failure paths leave the same process alive, MdmClientApplication.onCreate() will not run again to re-register keep-alive. The device is still installed and connected, but it has lost the normal TobService keep-alive recovery until some later process restart. The delayed 'result == 0 but still alive' recovery branch has the same issue: it clears retireInProgress without restoring keep-alive.

Please re-register pbsAppKeepAlive(packageName, true, 0) when retire recovery decides the uninstall did not actually remove the client, so a failed retire returns the device to the same recovery posture it had before the attempt.

hacha:
Fixed in the follow-up commit: every path where retire recovery concludes the client was not actually removed now calls a new `restoreKeepAlive()` — `failRetire()` (covering the failure callback and the thrown path) and the delayed "result 0 but still alive" branch — re-registering `pbsAppKeepAlive(packageName, true, 0)` best-effort before `retireInProgress` is cleared. The call is idempotent, so the binder-null path (where the keep-alive was never deregistered) is harmless. A failed retire now returns the device to the same recovery posture it had before the attempt: keep-alive immediately, guard via the next mutual-watch tick.

Verified: `:app:compileDevDebugKotlin` + `:app:testDevDebugUnitTest` green.

PR #60, opened 2026-07-12
feat: show server version, flag version drift, and offer a per-device client Update

hacha:
## Summary

Makes STYLY-MDM version state visible and actionable in the web console, and lets an operator bring an out-of-date device back in sync with one click.

Three cohesive changes:

1. **Server version in the console** — a new `SERVER_INFO` WebSocket message carries the running `styly_mdm` version; the console shows it next to the `STYLY-MDM` brand in the top bar.
2. **Version-drift highlighting** — compatibility is keyed on `major.minor` (the third "build" digit is reserved for compatible, independent updates, so a `v0.2.1` server and a `v0.2.0` client count as in sync). A stable-online client on an *older* `major.minor` than the server reddens that device's badge; a client *ahead* of the server reddens the top-bar server version instead. Offline / `updating` rows and the `0.0.0` untagged fallback are exempt. Comparison is numeric (`0.10` > `0.2`).
3. **Per-device Update button** — next to a behind device's red badge, when the server holds a client APK on the *same* `major.minor` as the server, an **Update** button installs it to that single device via the ordinary `INSTALL_APK` path, so the device drives the existing self-update handshake. The server finds its client APK by the `styly-mdm-client_<version>.apk` release naming convention in `APK_DIR`, and a new `CLIENT_APK_INFO` message keeps the console current (on connect and after every upload).

### How the server "holds" the client APK

CI (`publish-pypi.yml`) downloads the **signed** `styly-mdm-client_<tag>.apk` asset that `release.yml` builds (polling, since the two workflows run in parallel on the same release event) and drops it into `styly_mdm/client/` before the wheel build. It ships in the wheel via the `client/*` package-data glob, and `seed_bundled_client_apk()` copies it into the data dir on startup — so a released server holds a matching, same-signing-key client out of the box (a from-source run relies on an operator upload). The bundled APK must be the signed release build because Android rejects an update signed with a different key.

## Compatibility policy

Any change that requires the client and server to move together **must** bump the minor (or major) version; the build digit is for compatible updates only. Documented in `docs/DEVELOPMENT.md` and the README.

## Trade-off worth calling out

The CI step **couples the two release pipelines**: a failed client-APK build now blocks the server's PyPI publish (previously independent). This is deliberate fail-closed behavior — never publish a server that can't offer its client for self-update — but it is an operational change. It is reversible to fail-open (build the wheel without the APK plus a loud warning) if preferred.

## Review

Reviewed by Codex before opening; all three should-fix findings were addressed (no blocking issues):

- `CLIENT_APK_RE` now matches the timestamp collision suffix `unique_apk_path()` actually writes, so a re-uploaded client APK is detected (and the mtime tie-break prefers it).
- The Update button is offered only when the client APK is on the *same* `major.minor` as the server (was "≥ server"), so it never pushes a device past the server.
- CI downloads the *exact* `styly-mdm-client_<tag>.apk` asset (not a wildcard) and verifies that file exists.

## Verification

- `mdm-server`: **209 tests pass** (`env PYTHONPATH=mdm-server python -m pytest -q`), incl. new `tests/test_client_apk.py` (filename parsing, collision suffixes, numeric latest-selection, mtime tie-break, bundled-seed non-clobber, empty `CLIENT_APK_INFO`).
- Console `<script>` compiles under Node (`vm.Script`); version-compare, badge-color, and Update-button predicates verified against a case matrix.
- Live admin-WS end-to-end: message order `SERVER_INFO → CLIENT_APK_INFO → DEVICE_LIST → GROUP_LIST`, `CLIENT_APK_INFO` points at the newest client APK.
- `publish-pypi.yml`: YAML parses, bundle step precedes the wheel build, tag passed via `env` (injection-safe), `run` script `bash -n` clean.

### Not verified here (needs a real release / browser)

- The wheel actually bundling `client/*` is proven **by mechanism** (identical directive to the already-shipping `static/*`); a full local `build` was blocked only by an old local setuptools. Confirm on the first release that the Test PyPI wheel contains `styly_mdm/client/*.apk`.
- The CI download/fail-closed path only exercises on a real published release.
- The in-browser button click was not exercised (browser extension offline); the emitted HTML + CSS were checked.


hacha:
**Manual on-device verification ✅**

Tested the Update → self-update flow end-to-end on a real PICO device:

- A device on client `0.2.2-dev` (behind the `0.3.0` server) showed the red out-of-sync badge and an **Update** button.
- Clicking **Update** installed the server's `styly-mdm-client_v0.3.0.apk` (same debug signing key); the device self-updated — the console showed `updating` across the restart — and re-registered as `0.3.0-dev`.
- The badge returned to normal and the Update button disappeared once the device matched the server's `major.minor`.

This covers the flow the PR description listed under "Not verified here (needs a real release / browser)".

PR #71, opened 2026-07-20
fix: keep a reconnected device online when its stale socket closes (#70)

### LINKED ISSUE #70: A rebooted device stays offline in the console while its WebSocket is connected
## Symptom

After a device reboots, the console keeps showing it as **offline** even though the
client has already reconnected and registered. Operators see a rebooted headset
that "never comes back online", so the fleet looks half-dead before a session and
every command targeting online devices skips it.

Observed on DEV-001 (`PA94U0MGKA220311G`, client 0.3.0-dev) against the dev server:

| Observation | Result |
|---|---|
| Device-side TCP socket | `192.168.128.249 -> 192.168.128.214:7080` ESTABLISHED |
| Server-side TCP socket | same connection ESTABLISHED (server pid, fd 13) |
| Client log | no `going silent`; the connection window closed as CONNECTED |
| Console `DEVICE_LIST` | **offline** |

## What actually happens

`devices` is only ever written in two places: the `REGISTER` handler
(`styly_mdm/server.py:1183`) and the disconnect path (`styly_mdm/server.py:1480`).
The disconnect path keys on `device_id` alone and never checks that the socket
being torn down is the one currently registered:

```python
if device_id and device_id in devices:
    del devices[device_id]
```

A reboot does not always close the old socket cleanly, so the pre-reboot
connection can still be open server-side when the rebooted client connects and
re-registers. When the server's 30 s heartbeat finally declares the stale socket
dead, its teardown deletes the entry belonging to the **live** connection.

Reproduced deterministically against an isolated server instance:

```
after A registers:       online
after B registers:       online
after stale A closes:    offline   <- B is still open
B still open?            True
```

Timeline captured on hardware (reboot at 15:11:20):

```
15:11:47  rebooted client REGISTERs                  -> online
15:11:48  stale pre-reboot socket times out,
          its teardown deletes the live entry        -> offline
15:16:47  next BATTERY_UPDATE hits devices[device_id]
          (server.py:1241) -> KeyError, handler dies,
          socket drops, client reconnects            -> online again
```

So the device is falsely offline for one battery interval (~5 minutes) after
every affected reboot, and only recovers through an unhandled `KeyError` that
kills a healthy connection. If that recovery reconnect lands while the server is
unreachable, the client goes silent and the device stays offline indefinitely.

## Impact

- Rebooted devices are unusable from the console for ~5 minutes.
- `BATTERY_UPDATE` after eviction raises `KeyError` and tears down a working
  WebSocket; any in-flight install/push on that device dies with it.
- The same missing identity check sits above the delete: a stale teardown can
  also settle another connection's pending self-update / retire state.

Note: first reported on devices whose content app is set as the PICO business
home app, but the race is independent of that setting — those are simply the
devices that get rebooted and watched in operation.

## Approaches

1. Guard the disconnect path on socket identity: only tear down when
   `devices[device_id]["ws"] is ws`, keeping the `last_seen` stamp and the
   `DEVICE_LIST` broadcast inside that guard. Audit the pending self-update /
   retire branches above it for the same check.
2. Make `BATTERY_UPDATE` (and any other handler indexing `devices[device_id]`)
   tolerate a missing entry instead of raising.
3. Have the server close a device's previous connection when the same
   `device_id` registers again, so a duplicate never coexists.

hacha:
Fixes #70.

## Problem

A reboot does not always close the client's WebSocket cleanly, so the server can still
hold the pre-reboot connection open (it only notices at the next 30 s heartbeat) when
the rebooted client has already reconnected and re-registered. The disconnect path keyed
on `device_id` alone:

```python
if device_id and device_id in devices:
    del devices[device_id]
```

so the stale socket's teardown deleted the registration belonging to the **live**
connection. The console then showed a fully connected device as `offline` until its next
`BATTERY_UPDATE` hit `devices[device_id]`, raised `KeyError`, and killed the surviving
connection — an unhandled crash was the only recovery path, roughly 5 minutes later.

Observed on DEV-001 (`PA94U0MGKA220311G`): ESTABLISHED socket on both ends, no
`going silent` in the client log, `DEVICE_LIST` reporting `offline`, and the registry
showing a REGISTER and a disconnect 0.8 s apart.

## Change

`devices[device_id]["ws"]` is the connection every command is sent to, so it is now also
the only connection allowed to tear the entry down. Every per-connection teardown step in
`device_ws_handler` is gated on socket identity — deleting the registration, freeing
transfer slots, dropping the install-dispatch record, failing a pending self-update
verification — and `BATTERY_UPDATE` applies the same check, ignoring frames from a
superseded socket instead of raising.

No client-side change: the client was behaving correctly throughout.

## Verification

- `mdm-server/tests/test_duplicate_connection.py` — four aiohttp integration tests over
  loopback covering the eviction, the late-telemetry `KeyError`, the transfer slot, and a
  regression guard that a genuinely gone device still goes offline. Three of the four fail
  against the pre-fix handler.
- Full suite: 214 passed.

Not yet re-checked against the running dev server on hardware — that needs a server
restart, so the reboot-and-watch confirmation is left to the operator.

### DIFF
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -917,6 +917,25 @@ time a window opens, so a pushed config takes effect on the next network transit
 disconnect — no reinstall or reboot. A device that has never reached a server can only
 run on the defaults (the push channel needs a live socket).
 
+### Duplicate connections: the newest registration owns the device
+
+A reboot does not always close the client's socket cleanly — the server can still hold
+the pre-reboot WebSocket open (it only notices at the next 30 s heartbeat) when the
+rebooted client has already reconnected and re-registered. Two live connections then
+carry the same `device_id`, and the invariant is that **the newest `REGISTER` owns the
+entry in `devices`**: `devices[device_id]["ws"]` is the connection every command is sent
+to, and only that connection may tear the entry down.
+
+Every per-connection teardown step in `device_ws_handler` is therefore gated on socket
+identity (`devices[device_id]["ws"] is ws`) — deleting the registration, freeing
+transfer slots, dropping the install-dispatch record, and failing a pending self-update
+verification. Message handlers that index `devices[device_id]` apply the same check and
+ignore frames arriving on a superseded socket.
+
+Without that guard the stale socket's teardown deleted the *live* registration, so a
+reconnected device showed as `offline` in the console until its next `BATTERY_UPDATE`
+raised `KeyError` and killed the surviving connection (issue #70).
+
 ## Server Discovery Protocol
 
 STYLY-MDM supports automatic server discovery via UDP broadcast on the LAN.
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -926,11 +926,18 @@ carry the same `device_id`, and the invariant is that **the newest `REGISTER` ow
 entry in `devices`**: `devices[device_id]["ws"]` is the connection every command is sent
 to, and only that connection may tear the entry down.
 
-Every per-connection teardown step in `device_ws_handler` is therefore gated on socket
-identity (`devices[device_id]["ws"] is ws`) — deleting the registration, freeing
-transfer slots, dropping the install-dispatch record, and failing a pending self-update
-verification. Message handlers that index `devices[device_id]` apply the same check and
-ignore frames arriving on a superseded socket.
+`device_ws_handler` enforces that in two places, both via `_owns_device()`:
+
+- **Inbound frames.** One guard above the message dispatch chain drops every
+  device-originated message except `REGISTER` when the socket no longer owns the
+  device_id. Without it a leftover socket could release the live connection's transfer
+  slot (`DOWNLOAD_COMPLETE`), forward a terminal result (`INSTALL_RESULT`,
+  `PUSH_FILES_RESULT`), or consume a self-update verification — finishing the live
+  client's job behind its back. `REGISTER` is the one exception: that is how a
+  connection claims ownership in the first place.
+- **Teardown.** The `finally` block runs only for the owning connection: deleting the
+  registration, freeing transfer slots, dropping the install-dispatch record, and
+  failing a pending self-update verification.
 
 Without that guard the stale socket's teardown deleted the *live* registration, so a
 reconnected device showed as `offline` in the console until its next `BATTERY_UPDATE`
--- a/mdm-server/styly_mdm/server.py
+++ b/mdm-server/styly_mdm/server.py
@@ -1234,11 +1234,19 @@ async def device_ws_handler(request: web.Request) -> web.WebSocketResponse:
                     if battery is None:
                         log.warning("Invalid BATTERY_UPDATE from %s: %s", device_id, data)
                         continue
+                    # Telemetry from a socket the device has already replaced is not
+                    # worth an unhandled KeyError that takes the live connection's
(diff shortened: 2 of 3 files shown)


hacha:
Blocking issue: superseded device sockets can still mutate the live connection's job state through non-teardown result handlers.

The PR correctly gates disconnect cleanup and `BATTERY_UPDATE` on socket identity, but the other device-originated result paths still trust only `device_id`. In particular, `DOWNLOAD_COMPLETE` calls `release_transfer_slot(device_id, ...)` without checking that `devices[device_id]["ws"] is ws`:

- `mdm-server/styly_mdm/server.py:1267-1271`
- similarly for `INSTALL_RESULT` / `PUSH_FILES_RESULT` slot release and `VERIFY_APK_RESULT` self-update consumption at `mdm-server/styly_mdm/server.py:1436-1443`

I reproduced this against the PR head with two WebSockets registered as `dev0`: the install command was dispatched to the newer/live socket, then a `DOWNLOAD_COMPLETE` sent on the older/superseded socket emptied `pending_transfers` and produced `INSTALL_PROGRESS done=True` before the live client had reported completion.

That violates the new invariant documented by this PR: once the newer registration owns `devices[device_id]["ws"]`, older sockets must not be able to advance or settle per-device job state. Please apply the same current-socket guard to device result handlers that release transfer slots, forward terminal results, or consume self-update verification results.

hacha:
Good catch — confirmed and fixed in d7726fe.

You were right that the invariant was only half-enforced. Rather than adding the same check to each result handler, I replaced the two site-specific guards with a single ownership guard above the message dispatch chain, so **every** device-originated frame except `REGISTER` is dropped when the socket no longer owns the `device_id`. `REGISTER` stays exempt because that is how a connection claims ownership. Both the inbound guard and the teardown gate now go through one `_owns_device()` helper.

Two regression tests encode your reproduction:

- `test_e2e_download_complete_on_a_superseded_socket_does_not_release_the_slot` — the stale socket's `DOWNLOAD_COMPLETE` leaves `pending_transfers` intact and emits no `INSTALL_PROGRESS`, while the owning connection still drives the job to completion.
- `test_e2e_terminal_result_on_a_superseded_socket_is_not_forwarded` — a stale `INSTALL_RESULT` never reaches the console; the owning connection's result still does.

Both fail against the previous PR head and pass on the new one. Full suite: 216 passed.

Docs updated to describe the guard as two enforcement points (inbound frames, teardown) instead of teardown only.

hacha:
## Verified on hardware

Ran the branch against the dev server (`MDM_WS_PORT=7080`, `PYTHONPATH=mdm-server` so it loads the working tree) and rebooted DEV-001 (`PA94U0MGKA220311G`).

The race the fix targets occurred on this very reboot — the stale pre-reboot socket outlived the new registration by 2.5 s:

```
15:54:23  reboot issued
15:54:50  Device registered: PA94U0MGKA220311G (A94U0, client v7)
15:54:52  Superseded device connection closed, ignoring: PA94U0MGKA220311G
15:59:50  Battery update from PA94U0MGKA220311G: 100% charging=True
```

On the old code that 15:54:52 teardown is exactly what deleted the live registration. Console status polled every 10 s from 15:54:53 to 15:59:50 — **online for all 30 samples**, including across the 5-minute battery tick that previously raised `KeyError` and killed the surviving connection.

Before: offline for ~5 minutes after reboot, recovering only via an unhandled crash.
After: online from the post-reboot REGISTER onward, no gap.

PR #77, opened 2026-07-21
feat: reboot / power off a managed device from the console (#74)

### LINKED ISSUE #74: Remotely power off / reboot a managed device from the console
## Problem

There is currently no way to power off or reboot a managed device remotely from the MDM console. Once a PICO device is deployed (often in a location that is physically inconvenient to reach), an operator has no way to shut it down or restart it without walking up to the headset. This blocks common fleet-operation needs:

- **Power off** — put idle devices into a safe/off state at end of day, or before physical handling, without touching each headset.
- **Reboot** — recover a device that is misbehaving (frozen app, degraded state) or apply a clean restart after configuration changes, remotely.

Today the only power-related primitive in the codebase is an internal, self-update-recovery-only timed shutdown/startup (`PowerCycleTimers.kt`, PICO `openTimingShutdown` / `openTimingStartup`). It is not exposed as a server-initiated command and cannot be triggered on demand from the console.

## Desired behavior

From the console, an operator can:

- Trigger an **immediate power off** of one or more selected devices.
- Trigger a **reboot** of one or more selected devices.
- See the outcome (accepted / failed) reflected back in the console, consistent with how other device commands report results.

## Notes / context

The command path follows the existing WebSocket architecture:

- **Server**: admin request types are dispatched in `admin_ws_handler` (`mdm-server/styly_mdm/server.py:1561`), which fans out `EXECUTE_*` messages to the target device socket (e.g. `handle_launch_app`, `server.py:1915`).
- **Client**: incoming commands are dispatched in `MdmClientService.handleCommand` `when(type)` (`mdm-client/.../MdmClientService.kt:302`), and the PICO `IToBServiceProxy` proxy is already obtained/used for power timing in `PowerCycleTimers.kt`.

## Approaches (to be decided during implementation)

- Add `REBOOT_DEVICE` / `POWER_OFF_DEVICE` admin request types and corresponding `EXECUTE_REBOOT` / `EXECUTE_POWER_OFF` device commands, mirroring `LAUNCH_APP` → `EXECUTE_LAUNCH`, with `*_RESULT` responses.
- On the client, call the PICO Enterprise SDK (TobService) reboot / shutdown APIs via the existing `IToBServiceProxy`. Per PICO docs, device shutdown/reboot are **advanced / high-risk APIs** that require the `pico_advance_interface` meta-data flag in the manifest and are gated by device-owner status — this needs to be verified against our current manifest and device-owner setup.

## Open questions

- Should power-off / reboot be a single-device action, or support multi-select / group fan-out like install/launch?
- Is a confirmation step (and/or restricting to certain operators) warranted for a destructive action like power off?
- Does the deployed client already carry the `pico_advance_interface` high-risk permission, or does enabling this require a client update + redeploy?

hacha:
Closes #74.

## What

Remote **Reboot** and **Power Off** as a console fan-out action, mirroring `LAUNCH_APP`.

- **Server** (`server.py`): `REBOOT_DEVICE` / `POWER_OFF_DEVICE` admin requests → one `_dispatch_power_command` helper resolves targets and fans out payload-less `EXECUTE_REBOOT` / `EXECUTE_POWER_OFF` (un-throttled, launch-style), acking `REBOOT_SENT` / `POWER_OFF_SENT`. `REBOOT_RESULT` / `POWER_OFF_RESULT` added to the device→admin forward set.
- **Client** (`MdmClientService.kt`): `executePowerControl` calls `pbsControlSetDeviceAction(DEVICE_CONTROL_REBOOT | SHUTDOWN)` on the same uncast `IToBService` binder the silent-install path uses. No new manifest permission or SDK — `pico_advance_interface` is already declared.
- **Console** (`index.html`): one **Power Control** section (Reboot + Power Off) behind a shared confirmation checkbox + per-action `confirm()` dialog, mirroring Retire.
- **Docs** (`DEVELOPMENT.md`): protocol rows in all four tables + a "Remote Power Control" section.

## Outcome model

A successful reboot/shutdown tears down the socket before any `*_RESULT: success` could flush, so the client sends `*_RESULT: accepted` **before** invoking the SDK call and the real success signal is the **device dropping offline** (a reboot reconnects on its own after ~30–60s). `*_RESULT: fail` is only emitted when the SDK rejects the call (device stays up to report it).

## Power off while charging

Device testing surfaced that a PICO headset **refuses a full power-off while a USB/charging cable is connected** unless the device-wide setting is enabled — so the shutdown path calls `pbsControlSetPowerOffwithUSBCable(S_ON)` first. This is a **persistent, device-wide** setting: a cabled headset on a dock can be remotely powered off and then **stranded until it is physically woken** — the intended venue capability, gated behind the console confirmation. Reboot is unaffected (it cycles regardless of charge).

## Verification

- `222 passed` full `mdm-server` suite, incl. 6 new `test_power_control.py` (fan-out shape, `["*"]` wildcard, no-online → ERROR, unknown-id skip, and an E2E asserting `REBOOT_RESULT` is forwarded to admins `device_id`-stamped).
- `:app:compileDevDebugKotlin` clean — proves `pbsControlSetDeviceAction` and `pbsControlSetPowerOffwithUSBCable` resolve on `tobservicelib:4.3.16`.
- Console: `node --check` + a node harness against the real extracted `sendPower` / `logPowerResult`.
- **On device:** reboot verified; power off verified to require the USB-cable enable (this PR) for cabled devices — re-test after client redeploy.

No version bump here — that is a release-time `chore:` commit via `release-version-bump.yml`. At release this warrants a **MINOR** bump (0.4.0 → 0.5.0): both server and client changed.

### DIFF
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -250,6 +250,7 @@ STYLY-MDM/
 | `REGISTER` | Sent on connect. Fields: `device_id`, `model`, `ip`, `version_code` (integer), `version_name`, `startup_app` (optional) |
 | `BATTERY_UPDATE` | Battery telemetry. Fields: `device_id`, `level` (integer 0-100), `charging` (boolean), `timestamp` (epoch seconds) |
 | `LAUNCH_RESULT` | Result of an app launch. Fields: `status` (`success`/`fail`), `package_name`, `error` (optional) |
+| `REBOOT_RESULT` / `POWER_OFF_RESULT` | Acknowledgement of a power command. `status` is `accepted` (the client received it and flushed this before invoking the SDK — a successful reboot/shutdown tears down the socket first, so no `success` is ever sent) or `fail` (the SDK rejected the call, so the device stayed up to report it). Fields: `status`, `error` (optional). See [Remote Power Control](#remote-power-control). |
 | `INSTALL_RESULT` | Result of an APK install. Fields: `status` (`success`/`fail`), `apk_filename`, `result_code` (optional), `error` (optional) |
 | `DOWNLOAD_COMPLETE` | Sent right after a download finishes, before the local work it feeds (the install, or the unzip + mirror). Frees the server's transfer slot so the next queued device can start downloading. Fields: `task` (`install` / `push`; absent means `install`), `apk_filename` (install), `dest_path` + `delete_extras` (push). Optional — see the transfer-throttling note below. |
 | `PUSH_FILES_RESULT` | Result of a push or sync. Fields: `status` (`success`/`fail`), `dest_path`, `added`, `updated`, `deleted` (counts; `deleted` is always 0 for a push), `error` (optional) |
@@ -264,6 +265,7 @@ STYLY-MDM/
 | Message type | Description |
 |---|---|
 | `EXECUTE_LAUNCH` | Launch an app. Fields: `package_name`, `extra` |
+| `EXECUTE_REBOOT` / `EXECUTE_POWER_OFF` | Reboot or power off the device immediately via the PICO advanced device-control API (`pbsControlSetDeviceAction`). No fields. See [Remote Power Control](#remote-power-control). |
 | `EXECUTE_INSTALL` | Download and install an APK. Fields: `apk_url`, `apk_filename`, plus `full_sha256` + `cd_sha256` (reference hashes of the file being dispatched; present only when the APK is a local upload in `apks/`). The client verifies the download against `full_sha256` before installing; a **self**-update is refused outright when the hashes are absent. |
 | `EXECUTE_PUSH_FILES` | Download a bundle and apply it to a directory. Fields: `bundle_url`, `bundle_filename`, `dest_path`, `delete_extras` (boolean; `false` = copy/overwrite only, `true` = full mirror. Read with a `false` default — a missing field must never delete) |
 | `EXECUTE_VERIFY_APK` | Compute `size` + Central-Directory digest (plus diagnostics) for an installed package. Fields: `package_name` |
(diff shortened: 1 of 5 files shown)


hacha:
Review finding: power-off can report `accepted` before the charging precondition has actually succeeded.

In `MdmClientService.kt`, `executePowerControl()` sends and flushes `POWER_OFF_RESULT` with `status: "accepted"` before it attempts the power-off-with-USB setting:

- `sendPowerResult(resultType, "accepted", "")` / `awaitOutboundFlush(2_000)` happen first.
- `pbsControlSetPowerOffwithUSBCable(PBS_SwitchEnum.S_ON, 0)` runs afterward.
- If that setting call throws, the code only logs a warning and continues to `pbsControlSetDeviceAction(DEVICE_CONTROL_SHUTDOWN, ...)`.

That makes the console semantics weaker than the UI/docs promise. The console says power off works even while charging, and the docs explain that this setting is what makes a cabled headset actually power off. If the setting fails because of firmware, permission, or a transient TobService problem, the operator can see `accepted` while the cabled headset remains online.

Please treat `pbsControlSetPowerOffwithUSBCable(S_ON, 0)` as a local precondition for `POWER_OFF_RESULT: accepted`: run it before the accepted ack, and report `fail` if it cannot be applied. Reboot can keep the current pre-ack model since it does not depend on this setting.

hacha:
Good catch — fixed in 1b16bb8.

`pbsControlSetPowerOffwithUSBCable(S_ON, 0)` is now a **precondition** for the power-off path rather than a best-effort step after the ack:

- It runs **before** `POWER_OFF_RESULT: accepted` (and before the flush), so the ack now means "the charging prerequisite is in place and the device is going down."
- If it throws — its only failure signal, since the call returns `void` — the client reports `POWER_OFF_RESULT: fail` with the reason and **does not proceed** to `pbsControlSetDeviceAction`, instead of logging a warning and shutting down anyway.
- Reboot is unchanged: it keeps the plain pre-ack model since it does not depend on this setting.

The console already renders a `fail` result as a failure line, so an operator now sees `fail` (not `accepted`) when a cabled device could not be prepared to power off. `DEVELOPMENT.md` updated to describe the precondition-and-abort behavior.

PR #78, opened 2026-07-26
Support 64 GiB push bundles without blocking WebSockets

0suu:
## Summary

- raise the Push Files / Sync Folder bundle limit from 2 GiB to 64 GiB while keeping the APK limit at 2 GiB
- change the default transfer-slot timeout from 600 to 1800 seconds
- move bundle ZIP creation off the aiohttp event loop so large archives do not starve device/admin WebSocket heartbeats
- make cancellation cleanup safe and atomically reserve ZIP output names for concurrent same-name uploads
- document ZIP64, temporary disk-space, and timeout behavior

## Root cause

Bundle creation called Python `zipfile` synchronously from the aiohttp request handler. A 7 GB upload blocked the event loop during compression, causing both the admin console and device WebSockets to disconnect.

## Operational notes

The 30-minute timeout releases the server transfer slot but does not cancel a device download. A 64 GiB bundle needs about 92 minutes at 100 Mbps, so slow large transfers may be reported as timed out while continuing on the device. This behavior is documented in the developer guide.

Bundles above 4 GiB use ZIP64 and still require acceptance testing on the target PICO firmware. Worst-case temporary disk use can approach roughly 128 GiB on the server and 192 GiB on a device.

## Validation

- `uv run --extra dev pytest`: 230 passed
- `git diff --check`: passed
- restarted the development server and verified HTTP 200 plus device/admin WebSocket reconnection

### DIFF
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -414,12 +414,12 @@ STYLY-MDM/
 >    older clients that never emit `DOWNLOAD_COMPLETE`, and clients whose download
 >    failed outright).
 > 3. Device disconnect (frees every slot the device held, immediately).
-> 4. A per-device timeout (`MDM_TRANSFER_TIMEOUT` seconds, default **1800**) so a
+> 4. A per-device timeout (`MDM_TRANSFER_TIMEOUT` seconds, default **3600**) so a
 >    silent/stuck device cannot block the queue. Lowering it recovers stuck slots
 >    sooner but risks releasing a slow-but-healthy transfer early, which only
 >    relaxes throttling and never drops the job itself.
 >    The timeout does not cancel the device download. At 100 Mbps, a 64 GiB bundle
->    takes about 92 minutes ideally, so the default 30-minute timeout will mark that
+>    takes about 92 minutes ideally, so the default 60-minute timeout will mark that
 >    transfer failed and release its slot while the device continues downloading.
 >
 > `pending_transfers` is keyed by **`(device_id, task)`**, not by device: an admin can
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -390,7 +390,7 @@ STYLY-MDM/
 
 > **Transfer throttling.** An `INSTALL_APK` or a `PUSH_FILES` targeting a large group
 > would otherwise make every device pull the file from the server at the same instant
-> (an APK can be up to 2 GiB, a push bundle likewise), spiking LAN/server bandwidth.
+> (an APK can be up to 2 GiB and a push bundle 64 GiB), spiking LAN/server bandwidth.
 > Instead the server owns a pool of **N** transfer slots
 > (`MDM_MAX_CONCURRENT_TRANSFERS` env var / `--max-concurrent-transfers` flag,
 > default **5**) and every byte-moving fan-out — install, push, sync — must take one
@@ -414,10 +414,13 @@ STYLY-MDM/
 >    older clients that never emit `DOWNLOAD_COMPLETE`, and clients whose download
 >    failed outright).
 > 3. Device disconnect (frees every slot the device held, immediately).
-> 4. A per-device timeout (`MDM_TRANSFER_TIMEOUT` seconds, default **600**) so a
+> 4. A per-device timeout (`MDM_TRANSFER_TIMEOUT` seconds, default **1800**) so a
 >    silent/stuck device cannot block the queue. Lowering it recovers stuck slots
 >    sooner but risks releasing a slow-but-healthy transfer early, which only
 >    relaxes throttling and never drops the job itself.
+>    The timeout does not cancel the device download. At 100 Mbps, a 64 GiB bundle
+>    takes about 92 minutes ideally, so the default 30-minute timeout will mark that
+>    transfer failed and release its slot while the device continues downloading.
 >
 > `pending_transfers` is keyed by **`(device_id, task)`**, not by device: an admin can
 > push files to a group that is already installing an APK, so one device may hold an
@@ -474,6 +477,17 @@ STYLY-MDM/
 > |---|---|---|
 > | **Push Files** (file or folder) | `false` | Copy and overwrite. Nothing at the destination is removed; `deleted` is always 0. |
 > | **Sync Folder** (folder only) | `true` | Full mirror: extras at the destination — including now-empty directories — are deleted, so it ends up identical to the bundle (`rsync --delete` semantics). |
+>
+> The upload limit is **64 GiB of source files before compression**. There is no free-space
+> preflight: in the worst case the server briefly needs about 128 GiB for the reconstructed
+> tree plus ZIP, and a device may need about 192 GiB while it holds the ZIP, extracted staging
+> tree, and destination files. Python creates ZIP64 archives automatically when required, but
+> bundles above 4 GiB must be acceptance-tested on the target PICO firmware.
+>
+> ZIP creation runs through `asyncio.to_thread`; doing multi-gigabyte compression on aiohttp's
+> event-loop thread would starve device/admin WebSocket heartbeats. The worker is shielded from
+> request-task cancellation, and cancellation is not propagated until the worker exits, so
+> staging and partial-ZIP cleanup cannot race file reads or writes still running in the thread.
 >
 > These are two console panels rather than one panel with a delete toggle so the
 > destructive operation has to be chosen **by name**. Only Sync Folder shows the
--- a/mdm-server/styly_mdm/server.py
+++ b/mdm-server/styly_mdm/server.py
@@ -58,7 +58,7 @@ CLIENT_APK_RE = re.compile(r"^styly-mdm-client_v?(\d+(?:\.\d+)+)(?:-.*)?\.apk$")
 
 # Maximum number of device downloads allowed in flight across the whole server.
 # Fanning a file out to a large group otherwise makes every device pull it from the
-# server at the same instant (an APK can be up to 2 GiB, a push bundle likewise),
+# server at the same instant (an APK can be up to 2 GiB and a push bundle 64 GiB),
 # spiking LAN/server bandwidth and stalling transfers. Every byte-moving fan-out —
 # install, push, sync — gates its dispatch on the same pool of this many slots; a
 # slot frees as soon as its device reports the download finished. Read when the pool
@@ -70,7 +70,9 @@ MAX_CONCURRENT_TRANSFERS = max(1, int(os.environ.get("MDM_MAX_CONCURRENT_TRANSFE
 # default: a large APK over a slow LAN can legitimately take minutes. Lowering it
 # recovers stuck slots sooner at the risk of releasing a slow-but-healthy transfer
 # early, which only relaxes throttling and never drops the job itself.
-TRANSFER_TIMEOUT = float(os.environ.get("MDM_TRANSFER_TIMEOUT", "600"))
+DEFAULT_TRANSFER_TIMEOUT = 30 * 60
+TRANSFER_TIMEOUT = float(
+    os.environ.get("MDM_TRANSFER_TIMEOUT", str(DEFAULT_TRANSFER_TIMEOUT)))
 
 # The two kinds of byte-moving job. A device can be in one of each at the same time
 # (an admin can push files to a group already installing an APK), so both the slot
@@ -140,7 +142,7 @@ _apk_hash_cache: dict[tuple[str, int, int], dict] = {}
 # devices, which mirror it into a destination directory. Kept separate from APK_DIR so the
 # APK workflow is untouched.
 BUNDLE_DIR = DATA_DIR / "bundles"
-MAX_BUNDLE_SIZE = 2 * 1024 * 1024 * 1024  # 2 GiB (total uploaded bytes before zipping)
+MAX_BUNDLE_SIZE = 64 * 1024 * 1024 * 1024  # 64 GiB (total uploaded bytes before zipping)
 MAX_BUNDLE_ENTRIES = 5000  # cap files per bundle to bound tree reconstruction
 
 # Persistent per-device registry: serial -> {label, model, ip, last_seen, startup_app, battery}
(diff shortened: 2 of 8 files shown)


mdm-server/styly_mdm/server.py
```diff
@@ -140,7 +142,7 @@
 # devices, which mirror it into a destination directory. Kept separate from APK_DIR so the
 # APK workflow is untouched.
 BUNDLE_DIR = DATA_DIR / "bundles"
-MAX_BUNDLE_SIZE = 2 * 1024 * 1024 * 1024  # 2 GiB (total uploaded bytes before zipping)
+MAX_BUNDLE_SIZE = 64 * 1024 * 1024 * 1024  # 64 GiB (total uploaded bytes before zipping)
 MAX_BUNDLE_ENTRIES = 5000  # cap files per bundle to bound tree reconstruction
```
hacha, line 146: **Blocking:** the byte limit went up 32x but the file-count cap did not move.

Real push payloads (Unity `StreamingAssets`, Addressables output, a captured video set) hit 5000 files long before they hit 64 GiB, so in practice the new ceiling is only reachable with a handful of enormous files -- which is not the workload this PR is motivated by.

It is also a bad failure mode: the check at L2676 fires *during* the upload, so an operator sends many GiB before getting a 413.

Either raise this alongside `MAX_BUNDLE_SIZE`, or say plainly in the docs that 64 GiB is bounded by a 5000-entry cap. Right now the README promises something the server will not accept.

mdm-server/styly_mdm/server.py
```diff
@@ -1143,6 +1162,42 @@ def zip_tree(root: Path, dest_zip: Path) -> None:
                 zf.write(path, path.relative_to(root).as_posix())
 
 
+async def zip_tree_async(root: Path, dest_zip: Path) -> None:
```
hacha, line 1165: **Blocking (about `zip_tree`, which is just above and outside the diff):** it still uses `zipfile.ZIP_DEFLATED`.

Single-threaded zlib runs at roughly 40-60 MB/s, so a 64 GiB bundle spends about 20-30 minutes in pure compression *after* the upload has finished -- and the payloads this feature exists for (video, textures, AssetBundles) are already compressed, so deflate buys close to nothing for that cost. It also keeps peak temp disk near the documented 128 GiB instead of shrinking it.

`ZIP_STORED` (or `compresslevel=1`) on the bundle path would cut the wall clock by an order of magnitude and make the 60-minute transfer timeout much less likely to matter. I would want a measured number here before advertising 64 GiB.

Separately, a note on the executor: `asyncio.to_thread` uses the **default** loop `ThreadPoolExecutor`, which is the same pool aiohttp's `ZLibCompressor` offloads large-payload compression to. A multi-minute `zip_tree` occupies a worker for its whole run, so a few concurrent uploads can starve WebSocket compression. A dedicated `ThreadPoolExecutor` for bundle work would decouple the two -- and I suspect it is closer to the real cause of the symptom that `compress=False` is working around (see my comment on `admin_ws_handler`).

mdm-server/styly_mdm/server.py
```diff
@@ -1532,7 +1587,13 @@ async def device_ws_handler(request: web.Request) -> web.WebSocketResponse:
...
 
 async def admin_ws_handler(request: web.Request) -> web.WebSocketResponse:
-    ws = web.WebSocketResponse(heartbeat=30)
+    # Admin messages are small control JSON, so compression saves effectively
+    # nothing.  More importantly, Chrome/aiohttp interoperability after a long
+    # concurrent HTTP upload has produced reserved-bit errors on the browser's
+    # next control frame, closing this otherwise healthy connection.
+    # Keep compression enabled for the device channel, where messages can be
+    # larger, but remove the extension from this browser-facing channel.
+    ws = web.WebSocketResponse(heartbeat=30, compress=False)
```
hacha, line 1596: This is a symptom-level fix and the diagnosis in the comment does not hold for the aiohttp we actually run.

The comment blames Chrome/aiohttp interop producing reserved-bit errors, which points at permessage-deflate compressor-state corruption. But aiohttp 3.14 (what resolves here) already guards exactly that: `send_frame` in `aiohttp/_websocket/writer.py` takes `self._send_lock` for small compressed frames and wraps large ones in a shielded task, precisely so concurrent or cancelled sends cannot corrupt the compressor. Its `ZLibCompressor.compress` docstring spells the requirement out. So the mechanism the comment describes should already be closed.

Two alternatives worth ruling out before freezing this in:

- **Executor starvation.** `asyncio.to_thread(zip_tree, ...)` and aiohttp's compression offload share the default loop executor. A long zip holds a worker for minutes; enough concurrent uploads and compression queues behind it. `compress=False` makes the symptom disappear without addressing that, and the device channel at L1166 keeps the same exposure. A dedicated executor for zip work fixes the cause for both channels.
- **aiohttp version.** `pyproject.toml` allows `aiohttp>=3.9,<4`, and the locking above landed well after 3.9. If the failure was observed against an older aiohttp, raising the floor is the actual fix.

Also worth being explicit about: `test_admin_websocket.py` asserts `admin.compress == 0`, i.e. it verifies the setting, not that the failure this is meant to prevent is gone. Useful as a guard against someone flipping it back, but it should not be read as a regression test.

I am not against keeping `compress=False` -- admin traffic is small control JSON and loses nothing. I would just want the comment to say "workaround, cause not confirmed" rather than assert a mechanism, so nobody treats the underlying issue as closed.

mdm-server/styly_mdm/server.py
```diff
@@ -1098,6 +1100,23 @@ def unique_bundle_path(filename: str) -> Path:
...
+    """Atomically reserve a unique path before ZIP work leaves the event loop.
+
+    Once ZIP creation runs in a worker thread, another upload can arrive while the
+    first archive is still being built. Exclusively creating a placeholder prevents
+    both workers from selecting and writing the same destination.
+    """
+    while True:
+        destination = unique_bundle_path(filename)
+        try:
+            destination.touch(exist_ok=False)
```
hacha, line 1113: The reservation itself is right -- without it two concurrent uploads would race between `unique_bundle_path` and the write, which the move to a worker thread makes newly reachable.

One wrinkle: the placeholder is a valid-looking 0-byte `.zip` sitting in `BUNDLE_DIR` for the entire zip window (potentially tens of minutes now), and `BUNDLE_DIR` is served by `add_static("/bundles", ...)`. If the server is killed mid-zip it stays there forever, and nothing cleans it up on startup.

Reserving `<name>.zip.part` and `os.replace()`-ing it into place on success would give you the same atomic claim plus an artifact that is self-evidently incomplete and cannot be fetched under its final URL.

mdm-server/styly_mdm/server.py
```diff
@@ -2708,14 +2770,21 @@ async def upload_bundle_handler(request: web.Request) -> web.Response:
...
+        raise
     except Exception as e:
+        if bundle_path is not None:
+            bundle_path.unlink(missing_ok=True)
         log.exception("Failed to build uploaded bundle")
         return web.json_response({"error": f"Failed to build bundle: {e}"}, status=500)
     finally:
         shutil.rmtree(staging, ignore_errors=True)
 
+    assert bundle_path is not None
```
hacha, line 2787: `assert` is stripped under `python -O`, and then `bundle_path.stat()` on the next line raises `AttributeError` on `None`. Not a live risk today, but it is load-bearing for correctness rather than a debug check.

Moving `size = bundle_path.stat().st_size` inside the `try` -- or an explicit `if bundle_path is None: return 500` -- narrows the type without depending on assertions being enabled.

docs/DEVELOPMENT.md
```diff
@@ -414,10 +414,13 @@ STYLY-MDM/
 >    older clients that never emit `DOWNLOAD_COMPLETE`, and clients whose download
 >    failed outright).
 > 3. Device disconnect (frees every slot the device held, immediately).
-> 4. A per-device timeout (`MDM_TRANSFER_TIMEOUT` seconds, default **600**) so a
+> 4. A per-device timeout (`MDM_TRANSFER_TIMEOUT` seconds, default **3600**) so a
 >    silent/stuck device cannot block the queue. Lowering it recovers stuck slots
 >    sooner but risks releasing a slow-but-healthy transfer early, which only
 >    relaxes throttling and never drops the job itself.
+>    The timeout does not cancel the device download. At 100 Mbps, a 64 GiB bundle
+>    takes about 92 minutes ideally, so the default 60-minute timeout will mark that
```
hacha, line 422: **Important:** this paragraph documents that the flagship scenario the PR enables reports a false failure under default settings. 92 minutes of real transfer against a 60-minute timeout means the console says "failed" for a perfectly healthy 64 GiB push, every time.

The root of it is that one event carries two meanings: "release the slot so a stuck device cannot block the pool" and "mark this device's job as failed". The first is a fairness mechanism and firing it early is fine -- the doc already says so. The second is a user-visible outcome and firing it early is just wrong.

Options, roughly in order of how much I like them:

1. Split the two. Release the slot on timeout but leave the device in `transferring` until `DOWNLOAD_COMPLETE`, a terminal result, or a disconnect arrives.
2. Derive the timeout from bundle size instead of using one constant for a 10 MB push and a 64 GiB one.
3. Raise the default high enough to cover the advertised ceiling.

Documenting known-wrong behaviour is not the same as fixing it, and this is the part of the PR I would most want revisited before merge.

docs/DEVELOPMENT.md
```diff
@@ -475,6 +478,17 @@ STYLY-MDM/
 > | **Push Files** (file or folder) | `false` | Copy and overwrite. Nothing at the destination is removed; `deleted` is always 0. |
 > | **Sync Folder** (folder only) | `true` | Full mirror: extras at the destination — including now-empty directories — are deleted, so it ends up identical to the bundle (`rsync --delete` semantics). |
 >
+> The upload limit is **64 GiB of source files before compression**. There is no free-space
+> preflight: in the worst case the server briefly needs about 128 GiB for the reconstructed
+> tree plus ZIP, and a device may need about 192 GiB while it holds the ZIP, extracted staging
+> tree, and destination files. Python creates ZIP64 archives automatically when required, but
+> bundles above 4 GiB must be acceptance-tested on the target PICO firmware.
```
hacha, line 485: Two things this paragraph is honest about that I think should change the shipped limit rather than only the docs.

**Free space.** There is no preflight on either side. On the server a `shutil.disk_usage` check before accepting turns a mid-upload `ENOSPC` into an upfront error. On the device it matters more: `pushTempDir()` (`MdmClientService.kt:1124`) writes to `Downloads/styly-mdm/.push-tmp` on *shared* storage, so a failed large push eats user-visible space, and the `finally` cleanup does not run if the process is killed. A `StatFs` check before `downloadBundle` would be cheap. Note also that the 192 GiB worst case you document exceeds a PICO 4's total storage in both the 128 GB and 256 GB configurations, so the documented ceiling is not reachable on the target hardware at all.

**ZIP64.** Good call flagging this rather than assuming it works. Worth adding that the client already has an explicit ZIP64 rejection path -- `MdmClientService.kt:1458`, `throw IllegalStateException("zip64 not supported")`. That one is APK-integrity only and does not affect bundle extraction via `ZipFile`, but it does mean this codebase has never actually handled a ZIP64 archive on device.

Until a >4 GiB bundle has been extracted on target firmware, I would ship the ceiling at 4 GiB (or gate anything above 4 GiB behind an explicit confirmation) and raise it in a follow-up once the acceptance test exists. Shipping a documented-untested limit means the first person to hit it is an operator in the field, not us.

hacha (commented):
Reviewed at `d1b2f0d`. I checked the branch out and ran the suite: `uv run --extra dev pytest` -> **231 passed** (the description says 230 -- harmless drift, just noting it).

## Summary

Moving the synchronous ZIP work off the event loop is well done. The root cause is correctly identified, the double-shield in `zip_tree_async` genuinely handles repeated cancellation rather than just the first one, and `reserve_unique_bundle_path` closes a real race that the `to_thread` change would otherwise have opened. Those parts I would take as-is.

My concern is that the 32x limit increase (2 GiB -> 64 GiB) lands without any of the surrounding constraints moving with it. Details are inline; here is what I think blocks the merge.

### Blocking

1. **The description contradicts the code.** It says the transfer timeout goes "from 600 to **1800** seconds", but the code, the docs, and the tests all say **3600**. Please fix one or the other.
2. **`MAX_BUNDLE_ENTRIES` is still 5000** (inline). 64 GiB is only reachable with <=5000 very large files, which is not what a real content folder looks like.
3. **`zip_tree` still uses `ZIP_DEFLATED`** (inline). That is roughly 20-30 minutes of pure single-threaded compression for 64 GiB, on content that is already compressed.

### Important

4. The 60-minute timeout is knowingly wrong for the headline use case -- inline on `docs/DEVELOPMENT.md`.
5. `compress=False` on the admin socket is a symptom-level fix with an unverified diagnosis -- inline.
6. No free-space preflight on either side, and the device physically cannot hold 64 GiB -- inline.
7. ZIP64 on Android is untested. I would hold the ceiling at 4 GiB until it is -- inline.
8. **The console gives no feedback for a multi-GiB upload.** `ensureUploaded()` in `mdm-server/styly_mdm/static/index.html` (around L1448) is a plain `fetch` behind a static `'Uploading N file(s)...'` label: no progress, no cancel, no client-side size or entry-count precheck. At 2 GiB that was tolerable. At 64 GiB it is an hour-plus of a frozen label, and an over-limit selection is only rejected after every byte is already on the wire. A pre-POST total-size / file-count check is a few lines and should ship with this change.

### Minor

- `assert bundle_path is not None` -- inline.
- The 0-byte placeholder left by `reserve_unique_bundle_path` -- inline.
- `test_bundle_size_limit_is_64_gib` and `test_default_transfer_timeout_is_60_minutes` just restate the constants. They can only fail when someone deliberately changes a constant and forgets the test, so they carry little signal.
- `TRANSFER_TIMEOUT = float(\n    os.environ.get(...))` -- the hanging close-paren style differs from the rest of the file.
- `client_max_size` is `max(MAX_APK_SIZE, MAX_BUNDLE_SIZE) + 10 MiB`, so every POST route now accepts a 64 GiB body. Harmless today (both routes stream and enforce their own caps), but worth remembering the next time a route is added.

### Suggestion

The event-loop fix and the limit raise are both in `f874519`, which is the main reason this review has to weigh them together. Splitting them would help: the `to_thread` change is a clear win regardless of where the bundle ceiling lands, and it could go in while items 2, 3 and 7 are settled. `0f3c01d` is already independent and can be discussed on its own terms.

PR #82, opened 2026-07-29
Fix admin WebSocket disconnect after bundle uploads

0suu:
## Summary

- Disable WebSocket compression for the browser admin channel.
- Keep device WebSocket compression unchanged; device-side immunity is unverified.
- Add regression coverage for admin and device compression negotiation.
- Document the transport workaround.

## Why

The failure reproduced twice with 7 GB uploads on aiohttp 3.14.3. In both cases, after the upload completed, the server-side aiohttp reader rejected the first `PUSH_FILES` frame sent by Chrome on `/ws/admin` with `Received frame with non-zero reserved bits`.

This was observed on the inbound Chrome-to-server path. `/ws/device` uses the same reader for compressed frames sent by OkHttp, so it is not proven immune and remains under observation. Admin messages are small, so disabling compression only on `/ws/admin` avoids the observed failure with negligible overhead.

## Test

- `uv run --extra dev pytest` — 223 passed

### DIFF
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -243,6 +243,10 @@ STYLY-MDM/
 
 ## WebSocket Protocol Reference
 
+Transport note: `/ws/admin` intentionally does not negotiate per-message
+compression as a workaround for the Chrome-to-server reserved-bit failure
+documented in PR #82. `/ws/device` keeps compression enabled for device traffic.
+
 ### Device → Server
 
 | Message type | Description |
--- a/mdm-server/styly_mdm/server.py
+++ b/mdm-server/styly_mdm/server.py
@@ -1532,7 +1532,13 @@ async def device_ws_handler(request: web.Request) -> web.WebSocketResponse:
 # ---------------------------------------------------------------------------
 
 async def admin_ws_handler(request: web.Request) -> web.WebSocketResponse:
-    ws = web.WebSocketResponse(heartbeat=30)
+    # Admin messages are small control JSON, so compression saves effectively
+    # nothing. More importantly, Chrome/aiohttp interoperability after a long
+    # concurrent HTTP upload can produce reserved-bit errors on the browser's
+    # next control frame, closing this otherwise healthy connection.
+    # Keep compression enabled for the device channel, where messages can be
+    # larger, but remove the extension from this browser-facing channel.
+    ws = web.WebSocketResponse(heartbeat=30, compress=False)
     await ws.prepare(request)
     admin_connections.add(ws)
     log.info("Admin console connected from %s", request.remote)
--- a/mdm-server/styly_mdm/server.py
+++ b/mdm-server/styly_mdm/server.py
@@ -1532,10 +1532,10 @@ async def device_ws_handler(request: web.Request) -> web.WebSocketResponse:
 # ---------------------------------------------------------------------------
 
 async def admin_ws_handler(request: web.Request) -> web.WebSocketResponse:
+    # A compressed admin connection was observed closing with a reserved-bit
+    # protocol error on the first control frame after a long HTTP upload.
     # Admin messages are small control JSON, so compression saves effectively
-    # nothing. More importantly, Chrome/aiohttp interoperability after a long
-    # concurrent HTTP upload can produce reserved-bit errors on the browser's
-    # next control frame, closing this otherwise healthy connection.
+    # nothing on this channel.
     # Keep compression enabled for the device channel, where messages can be
     # larger, but remove the extension from this browser-facing channel.
     ws = web.WebSocketResponse(heartbeat=30, compress=False)
--- a/mdm-server/styly_mdm/server.py
+++ b/mdm-server/styly_mdm/server.py
@@ -1532,8 +1532,11 @@ async def device_ws_handler(request: web.Request) -> web.WebSocketResponse:
 # ---------------------------------------------------------------------------
 
 async def admin_ws_handler(request: web.Request) -> web.WebSocketResponse:
-    # A compressed admin connection was observed closing with a reserved-bit
-    # protocol error on the first control frame after a long HTTP upload.
+    # On aiohttp 3.14.3, the server-side reader rejected Chrome's first PUSH_FILES
+    # text frame after a long upload with "Received frame with non-zero reserved
+    # bits". The failure was observed only on the inbound Chrome admin frame.
+    # The device channel receives compressed OkHttp frames through the same
+    # reader, so device-side immunity is unverified and remains under observation.
     # Admin messages are small control JSON, so compression saves effectively
     # nothing on this channel.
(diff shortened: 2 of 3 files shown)


hacha:
Reviewed — the change looks good to merge. Verification I ran locally:

- Checked out the branch and ran the full suite: **223 passed**.
- Reverted `compress=False` and re-ran the new test: it fails (`admin.compress == 15`), so the regression guard genuinely exercises the fix.
- The fix itself is minimal and correctly scoped to the admin handler.

A few comments, none blocking:

**1. One question that settles both the root cause and the device-channel scope: where exactly was the reserved-bit error observed — in Chrome's console, or as a server-side aiohttp `WSProtocolError`?**

This matters because the reported scenario ("first control frame after a long bundle upload") is *also* the device-channel scenario: right after an upload, the server sends `EXECUTE_PUSH_FILES` over `/ws/device`. If the corruption is on the server's sending side, the device channel is equally exposed and this fix is under-scoped. If it's a Chrome-side receive issue, devices (OkHttp) are unaffected and keeping compression there is fine. Right now the code comment asserts the latter without stating the evidence — please record the exact error text and where it was captured in the PR description, and make the comment say either "Chrome-specific, verified" or "device side unverified, watching".

A reserved-bit error means the *receiver* saw an unexpected RSV bit. aiohttp 3.14.x's writer never sets RSV on control frames and writes each frame without awaiting mid-frame (I read `_websocket/writer.py` to confirm), so a receive-side frame-boundary misread after the large preceding frame is the more plausible mechanism — which would also explain why `compress=False` helps (no RSV1 frames left to misread). Related but probably not the cause: aiohttp 3.14.2 shipped a permessage-deflate fix for control frames interleaved into fragmented messages (aio-libs/aiohttp#12988); its symptom profile doesn't match, but it's worth confirming which aiohttp version the failing server ran, and that the failure still reproduces on 3.14.3 before treating `compress=False` as permanent.

**2. Test comment overstates what the test does.** "Exercise the same browser-to-server command that followed a multi-gigabyte upload" — the test reproduces neither the upload, the size, nor the concurrency. The essential assertions are the two `compress` checks; the ~50-line PUSH_FILES flow duplicates `test_push_files.py`. Either trim the test down to the negotiation checks, or reword the comment to match reality.

**3. Docs:** please add a line to the WebSocket Protocol Reference section of `docs/DEVELOPMENT.md` noting the admin channel is intentionally uncompressed, so a future refactor doesn't silently re-enable it.

(Minor: PR body says 226 passed, I got 223 locally — unverified why, likely environment/branch drift; flagging for completeness only.)

0suu:
1.Server-side. The failure reproduced twice with 7 GB uploads on aiohttp 3.14.3. In both cases, after the upload completed, aiohttp's server-side reader rejected the first PUSH_FILES frame sent by Chrome on /ws/admin with Received frame with non-zero reserved bits.
This was on the inbound Chrome-to-server path. /ws/device uses the same reader for compressed frames sent by OkHttp, so the device channel is not proven immune. I updated the code comment and PR description to state that the device side is unverified and remains under observation.

2.Agreed. I trimmed the test to the admin/device compression negotiation checks and removed the claim about reproducing the large-upload flow. I also restored a comment explaining why device compression is asserted.

3.Agreed. I added a transport note explaining that /ws/admin is intentionally uncompressed as a workaround documented in PR #82, while /ws/device keeps compression enabled.

@hacha

hacha:
Thanks — the trimmed test, the updated code comment, and the docs note all look right, and the version detail in your reply turned out to be the missing piece. **Still approve to merge.** But it identifies the root cause, and it changes the device-channel answer from "unverified" to "affected".

Correcting my earlier comment: I raised aiohttp #12988 and then dismissed it as "probably not the cause". The direction was right and my sign was wrong — **#12988's own fix introduced the regression you hit.**

## Root cause: aiohttp #13274

[aio-libs/aiohttp#13274](https://github.com/aio-libs/aiohttp/issues/13274) (filed 2026-07-27):

> aiohttp **3.14.2/3.14.3** rejects a valid compressed WebSocket data frame when the first inbound frame is a control frame.
> 1. Server sends PING → 2. client replies PONG → 3. client sends a compressed TEXT frame → 4. server closes with 1002: `Received frame with non-zero reserved bits`
> The same sequence works correctly with aiohttp **3.14.1**.

Error string, affected versions, inbound direction, and trigger all match your report exactly. It also explains the "after a long upload" framing: a long upload leaves the admin channel idle, a heartbeat PING/PONG lands in that gap, and the next compressed frame is the one that gets rejected.

| Version | Date | State |
|---|---|---|
| 3.14.1 | 06-07 | Good (this is what our venv has — why the suite passes locally and I couldn't reproduce) |
| 3.14.2 | 07-20 | #12988 landed → regression introduced |
| 3.14.3 | 07-22 | Still affected (what you reproduced on) |
| — | 08-03 | Fix [#13302](https://github.com/aio-libs/aiohttp/pull/13302) merged to the 3.14 branch — **not yet released** |

## The device channel is affected, not just unverified

The bug fires on "compressed data frame preceded by an inbound control frame". All three preconditions hold on `/ws/device`:

1. **OkHttp negotiates compression** — 4.11.0 adds `Sec-WebSocket-Extensions: permessage-deflate` unconditionally (verified in the jar's `RealWebSocket` bytecode).
2. **We send frames over the compression threshold** — `minWebSocketMessageToCompress` defaults to 1024 and we don't override it (`WebSocketManager.kt:68`, verified in `OkHttpClient$Builder` bytecode). `VERIFY_DIR_RESULT` carries up to 2,000 `{relative_path, size, sha256}` entries (`MdmClientService.kt:1620`); a 64-char hash per entry means roughly a dozen files clears 1 KB.
3. **Inbound control frames are constant** — client `pingInterval` 15 s (`WebSocketManager.kt:70`) plus server `heartbeat=30` (`server.py:1166`).

To be precise about what I did and didn't do: I verified the OkHttp behavior from bytecode and the message shape from source, but **I did not observe a device disconnect** on 3.14.2/3.14.3. This is "expected to fail", not "observed failing". Cheap confirmation if you want it: take the reproducer in #13274, point `ws_connect(..., compress=15)` at `/ws/device`, and send a >1 KB payload after a PING/PONG on 3.14.3.

## Suggested follow-up: keep `compress=False`, add a version exclusion

```
dependencies = ["aiohttp>=3.9,!=3.14.2,!=3.14.3,<4"]
```

Additive, not a replacement:

- 3.14.4 doesn't exist yet, so **the exclusion is currently the only thing that protects `/ws/device`** — the admin-side `compress=False` does nothing for it.
- `compress=False` is still worth keeping on its own merits: compression buys nothing on the browser channel, and it holds regardless of what upstream does next.

Fold this into the PR or take it separately, whichever you prefer — no need to grow a one-line workaround into a dependency-policy change. Worth updating the code comment either way, since "device-side immunity is unverified" now understates it.

hacha (approved):
LGTM

PR #95, opened 2026-07-29
feat: implement durable Push / Sync jobs (#91)

### LINKED ISSUE #91: Define push job identity, lifecycle, and capability negotiation
Parent: #83

## Summary

Introduce a server-owned push job model before adding byte-level progress, cancellation, inactivity detection, or large-payload admission.

The current push protocol is not job-scoped. `PUSH_PROGRESS`, `PUSH_DEVICE_STATE`, `DOWNLOAD_COMPLETE`, and `PUSH_FILES_RESULT` cannot distinguish concurrent or retried jobs. The server also keys pending push transfers only by device and task, while the console keeps only one current push-progress entry.

This issue defines the canonical identity, state machine, compatibility contract, and recovery snapshot consumed by the dependent progress and cancellation work.

### Job identity

- Allocate an opaque server-owned `job_id` before browser-to-server upload starts.
- Add a lightweight job-creation endpoint or equivalent handshake.
- Carry `job_id` through upload/packaging records and `PUSH_FILES_SENT`, `PUSH_PROGRESS`, `PUSH_DEVICE_STATE`, `EXECUTE_PUSH_FILES`, `DOWNLOAD_COMPLETE`, and `PUSH_FILES_RESULT`.
- Give each job snapshot/event a monotonically increasing `revision`.
- A stale or mismatched result must not release another job's slot or mutate its state.

### Canonical lifecycle

Define and document server-owned job states covering creation, upload, packaging, ready, running, reconciliation, and terminal outcomes.

Define per-device states covering queued, downloading, validating, applying, reconciliation, and terminal outcomes.

The following invariants are required:

- State transitions are monotonic except for explicit retry/reconciliation.
- Transfer-slot release is not a terminal device or job state.
- A job is terminal only after every selected device is terminal, or after failure before dispatch.
- Aggregate counts are derived from canonical per-device state.
- At most one Push/Sync applies on a device at a time; overlapping work is queued or rejected explicitly.

### Capability negotiation

- Extend device `REGISTER` with an optional `capabilities` array.
- Define `push_job_id_v1`; capable clients accept and echo `job_id`.
- Reserve `push_progress_v1` for #85.
- Reserve `push_resume_v1` for #94; capable clients report and resume validated artifact offsets.
- Determine support from capabilities, not inferred version numbers.
- Missing or malformed capabilities mean no optional support.
- New clients remain compatible with old servers when `job_id` is absent.
- Legacy clients remain supported under an explicit one-active-push-per-device restriction.
- Operations requiring recoverable state reject unsupported targets before upload and report them clearly.

### Reload and restart recovery

- Return active-job and bounded recent-job snapshots when an admin WebSocket connects.
- Include `job_id`, `revision`, job/per-device states, timestamps, destination, mode, failures, immutable artifact identity, current attempt, and validated resumable offsets.
- Reconstruct the console after reload from the server snapshot.
- Persist enough metadata atomically that server restart does not silently forget accepted jobs.
- Define restart reconciliation:
  - non-resumable upload/packaging becomes `interrupted`;
  - dispatched devices enter reconciliation and may report validated resumable state;
  - matching late results may settle during a bounded window;
  - commands are not automatically re-dispatched.

## Acceptance criteria

- [ ] Concurrent jobs, including identical bundle/destination labels, have distinct IDs and state.
- [ ] All capable lifecycle messages carry the correct `job_id`.
- [ ] Snapshots/events carry monotonic revisions; stale revisions are ignored.
- [ ] Stale or mismatched results cannot release or complete another job.
- [ ] Pushes targeting the same device are serialized or explicitly rejected.
- [ ] Finishing downloads does not complete a job while validation/apply is active.
- [ ] Console reload restores active and recent job state.
- [ ] Server restart behavior is covered for upload, packaging, download, and apply.
- [ ] Legacy clients retain the documented small-push flow and restrictions.
- [ ] Unsupported targets are reported before starting work requiring new capabilities.
- [ ] Tests cover concurrent jobs, stale/late results, duplicate connections, reload, restart, and legacy clients.
- [ ] Development documentation and message-flow diagrams are updated.

## Non-goals

- Byte-level progress UI; covered by #85.
- Cancellation; covered by #89.
- Inactivity policy and stale-artifact recovery; covered separately.
- Resumable browser-to-server upload.
- Automatic redispatch after restart; server-authorized #94 reconciliation resume remains allowed.

0suu:
## Summary

Implements #91: durable, server-owned Push / Sync jobs.

## Included

- Create-before-upload `job_id` and idempotent `client_request_id`
- Immutable artifacts with exact size and SHA-256 identity
- SQLite-backed job and per-device lifecycle, queueing, and transfer ownership
- Durable Android active state, terminal-result outbox, ACK handling, and reconciliation
- Restart recovery with persistent device fences; queued work is never automatically redispatched
- Revisioned Admin snapshots and concurrent-job console rendering
- Restricted legacy-client and legacy-server compatibility

## Safety properties

- Lifecycle messages use exact job identity, so stale work cannot mutate another assignment or release its transfer slot.
- The client verifies artifact size and SHA-256 before extraction or apply.
- The 2 GiB limit applies to uploaded source bytes and extracted content; ZIP container overhead is accepted.
- Network download completion is distinct from validation, apply, and terminal job completion.

## Not included

Byte-progress UI, cancellation, upload resume, HTTP Range resume, and Install job-model integration.

## Review guide

1. `push_jobs.py` and `push_job_store.py` — identity, state machine, transactions, and snapshots
2. `push_job_manager.py`, `push_scheduler.py`, and `push_runtime.py` — queueing, ownership, recovery, and HTTP/WebSocket integration
3. Android `PushProtocol`, `PushJobCoordinator`, `PushJobStore`, and `PushFilesWorker`
4. Console adapter, documentation, and regression tests

Closes #91


hacha:
### Code review

Verified locally at `141cc74`: `mdm-server` `pytest` 355 passed, `mdm-client` `gradlew testDebugUnitTest` passed. The four safety properties in the description hold as written — SHA-256 and exact size are enforced before extraction ([`PushProtocol.kt#L158`](https://github.com/styly-dev/STYLY-MDM/blob/141cc74b252df654ec0fe051cfb60c4717177def/mdm-client/app/src/main/java/com/styly/mdmclient/PushProtocol.kt#L158) makes both mandatory for job-v1), the 2 GiB bound is applied to both declared and actual uploaded bytes, transfer slots are released only by exact `(job_id, device_id, attempt)` identity, and legacy clients keep working (`MDM_ALLOW_LEGACY_PUSH` defaults to `1`). Transaction discipline is sound: `BEGIN IMMEDIATE` + WAL + `synchronous=FULL` on a single dedicated DB thread, with `_commit`/`_rollback` on every exit path.

Found 5 issues.

**1. A device that goes offline mid-push is permanently excluded from all future jobs.**

[`push_job_manager.py#L756-L758`](https://github.com/styly-dev/STYLY-MDM/blob/141cc74b252df654ec0fe051cfb60c4717177def/mdm-server/styly_mdm/push_job_manager.py#L756-L758) computes

```python
replaced = (old_process is not None and old_process != process_instance_id) \
        or (old_process is None and protocol is ProtocolMode.LEGACY)
```

A **job_v1** fence with a NULL `blocking_process_instance_id` matches neither disjunct and can never be auto-cleared. That NULL is the common case, not an edge case: [`push_runtime.py#L1719-L1725`](https://github.com/styly-dev/STYLY-MDM/blob/141cc74b252df654ec0fe051cfb60c4717177def/mdm-server/styly_mdm/push_runtime.py#L1719-L1725) passes `session.process_instance_id if session else None`, and `mark_unconfirmed` is invoked from `_reconciliation_housekeeping` exactly when the reconciliation deadline (`MDM_PUSH_RECONCILIATION_TIMEOUT`, default 1800 s) elapsed — i.e. when the device is offline and `session is None`.

Sequence: device goes offline mid-push → 30 min later a fence is written with `protocol_mode='job_v1'`, `blocking_process_instance_id=NULL` → the device reconnects as a brand-new process (even after a factory reset) → `clear_fence_on_process_replacement` returns `[]` forever → `claim_next`'s `NOT EXISTS push_device_fences` excludes that device from every subsequent job. No automatic reconcile rescues it: [`push_runtime.py#L768-L771`](https://github.com/styly-dev/STYLY-MDM/blob/141cc74b252df654ec0fe051cfb60c4717177def/mdm-server/styly_mdm/push_runtime.py#L768-L771) sets `needs_reconcile` only for `RECONCILING`, not `UNCONFIRMED`. Recovery requires a manual console **Reconcile**.

[`docs/PUSH_JOBS.md`](https://github.com/styly-dev/STYLY-MDM/blob/141cc74b252df654ec0fe051cfb60c4717177def/docs/PUSH_JOBS.md#L84) states the intended rule ("a new job-v1 process also safely replaces a legacy fence that had no process UUID"); the offline-timeout case falls through both halves of it. The existing test at `test_push_job_store.py:404` only covers the online case with a non-null `"process-a"`.

**2. A device cell freezes in the console when a second Push is queued onto a busy device.**

[`push-jobs-v1.js#L448-L452`](https://github.com/styly-dev/STYLY-MDM/blob/141cc74b252df654ec0fe051cfb60c4717177def/mdm-server/styly_mdm/static/push-jobs-v1.js#L448-L452) discards `applyAssignment`'s boolean return and unconditionally runs `renderedAssignments.set(deviceId, selected.job.job_id)`, while [`index.html#L1348-L1353`](https://github.com/styly-dev/STYLY-MDM/blob/141cc74b252df654ec0fe051cfb60c4717177def/mdm-server/styly_mdm/static/index.html#L1348-L1353) refuses the apply when `current.pending_request_id !== assignment.client_request_id`.

D1 is mid-transfer for job A (request R1); the operator starts a second push targeting D1, so [`index.html#L1625-L1629`](https://github.com/styly-dev/STYLY-MDM/blob/141cc74b252df654ec0fe051cfb60c4717177def/mdm-server/styly_mdm/static/index.html#L1625-L1629) optimistically overwrites the cell with `{job_id: null, pending_request_id: R2}`. Every later snapshot still selects job A (active outranks queued), but `applyAssignment(A)` returns false from then on. D1's cell stays frozen on the optimistic queued state for the rest of job A, and A's terminal result for D1 is never painted — there is no fallback painter, since the server emits no `PUSH_DEVICE_STATE` for job-v1 devices. The optimistic cell also has `job_id: null`, so `clearAssignment` (which requires `current.job_id === jobId`) can never remove it. Server state stays correct and a reload repaints, but the operator is shown a stale cell.

**3. `PUSH_FILES` is dispatched on a socket captured before the upload.**

[`push-jobs-v1.js#L148`](https://github.com/styly-dev/STYLY-MDM/blob/141cc74b252df654ec0fe051cfb60c4717177def/mdm-server/styly_mdm/static/push-jobs-v1.js#L148) calls `nativeSend.call(socket, ...)`, where `socket` was captured before `POST /api/push-jobs` and before an artifact upload that runs for minutes on a large bundle. Lines 294 and 343 in the same file correctly use `currentAdminSocket`. If the admin socket drops during the upload (index.html reconnects automatically), `send()` on a CLOSED socket throws `InvalidStateError`, and the catch at line 151 reports "Push/Sync job failed before dispatch" — while the job is fully created, uploaded and `ready` server-side.

**4. A device-scoped fence is rendered as if it belonged to each job.**

[`push_job_store.py#L387`](https://github.com/styly-dev/STYLY-MDM/blob/141cc74b252df654ec0fe051cfb60c4717177def/mdm-server/styly_mdm/push_job_store.py#L387) resolves `device_fence` with `WHERE device_id=?` and embeds the result in every job's device row. Once D1 is fenced by a new job, every job that ever targeted D1 — including long-completed `succeeded` jobs — gains "fenced D1 (blocked by \<other-job\>)" and a Reconcile button in [`push-jobs-v1.js#L268`](https://github.com/styly-dev/STYLY-MDM/blob/141cc74b252df654ec0fe051cfb60c4717177def/mdm-server/styly_mdm/static/push-jobs-v1.js#L268). This misattributes the block and re-decorates finished history.

**5. `StoreConflict` is retried without backoff.**

[`push_scheduler.py#L444-L446`](https://github.com/styly-dev/STYLY-MDM/blob/141cc74b252df654ec0fe051cfb60c4717177def/mdm-server/styly_mdm/push_scheduler.py#L444-L446) — in `_recover_unexpected_dispatch_failure`, the generic `except Exception` branch sleeps `_RUN_RETRY_DELAY`, but `except StoreConflict: continue` has no delay and no attempt cap. Each iteration awaits the DB worker so the event loop is not blocked, but a persistently conflicting row becomes a hot loop against the single DB thread.

---

Two notes outside the diff, for follow-up rather than this PR:

- `PushJobStore.kt`'s `trim()` bounds `completedReceipts` but leaves `pendingResults` unbounded, and `replayPendingResults()` re-sends the entire list on every `REGISTERED`. If ACKs are lost, a frequently reconnecting device sends a linearly growing burst per reconnect and `state.json` grows without bound.
- `.github/workflows/` has no `on: pull_request` or `on: push` workflow, so `gh pr checks` reports nothing for this branch and neither test suite runs automatically. A +12,766-line change to the concurrency and persistence core is a strong argument for adding one.

Overall this is unusually disciplined work for its size — single-threaded SQLite ownership, identity-checked waiters, explicit reasoning for each ordering decision, and `PUSH_JOBS.md` documenting the deliberate deviations from the issue's proposed layout. Issue 1 is the one I would treat as a blocker.

mdm-server/styly_mdm/push_job_manager.py
```diff
@@ -0,0 +1,1168 @@
...
+                    self.store._commit(conn)
+                    return []
+                old_process = fence["blocking_process_instance_id"]
+                protocol = ProtocolMode(fence["protocol_mode"])
+                # A different recorded process proves replacement.  For a legacy fence
+                # no process UUID existed; a new job-v1 process is itself the safe
+                # migration evidence required by §19.3.
+                replaced = (
+                    old_process is not None and old_process != process_instance_id
+                ) or (old_process is None and protocol is ProtocolMode.LEGACY)
```
hacha: **Blocker: a device that goes offline mid-push is permanently excluded from all future jobs.**

A **job_v1** fence whose `blocking_process_instance_id` is NULL matches neither disjunct here, so it can never be auto-cleared. That NULL is the common case, not an edge case: `push_runtime.py:1721` passes `session.process_instance_id if session else None`, and `mark_unconfirmed` is invoked from `_reconciliation_housekeeping` exactly when the reconciliation deadline (`MDM_PUSH_RECONCILIATION_TIMEOUT`, default 1800 s) elapsed — i.e. when the device is offline and `session is None`.

Sequence:

1. Device goes offline mid-push.
2. 30 min later a fence is written with `protocol_mode='job_v1'`, `blocking_process_instance_id=NULL`.
3. The device reconnects as a brand-new process (even after a factory reset).
4. `clear_fence_on_process_replacement` returns `[]` forever.
5. `claim_next`'s `NOT EXISTS push_device_fences` excludes that device from every subsequent job.

No automatic reconcile rescues it either: `push_runtime.py:768` sets `needs_reconcile` only for `RECONCILING`, not `UNCONFIRMED`. Recovery requires a manual console **Reconcile**.

`docs/PUSH_JOBS.md:84` states the intended rule ("a new job-v1 process also safely replaces a legacy fence that had no process UUID") — the offline-timeout case falls through both halves of it. The existing test at `test_push_job_store.py:404` only covers the online case with a non-null `"process-a"`.

Suggestion: treat a fresh job-v1 process UUID as replacement evidence when `old_process is None`, regardless of the fence's `protocol_mode`, and/or include `UNCONFIRMED` in the `needs_reconcile` condition so reconnect triggers an automatic exact reconcile.

mdm-server/styly_mdm/static/push-jobs-v1.js
```diff
@@ -0,0 +1,510 @@
...
+      Object.keys(job.devices || {}).forEach(function (deviceId) {
+        deviceIds.add(deviceId);
+      });
+    });
+    deviceIds.forEach(function (deviceId) {
+      const selected = selectedAssignmentFor(deviceId);
+      const renderedJobId = renderedAssignments.get(deviceId);
+      if (selected) {
+        bridge.applyAssignment(assignmentView(selected, deviceId));
+        renderedAssignments.set(deviceId, selected.job.job_id);
```
hacha: **A device cell freezes in the console when a second Push is queued onto a busy device.**

`applyAssignment`'s boolean return is discarded, and `renderedAssignments.set(...)` runs unconditionally. But `index.html:1348-1353` refuses the apply when `current.pending_request_id !== assignment.client_request_id`.

Scenario: D1 is mid-transfer for job A (request R1); the operator starts a second push targeting D1, so `index.html:1625-1629` optimistically overwrites the cell with `{job_id: null, pending_request_id: R2}`. Every later snapshot still selects job A (active outranks queued), but `applyAssignment(A)` returns false from then on. D1's cell stays frozen on the optimistic queued state for the rest of job A, and A's terminal result for D1 is never painted — there is no fallback painter, since the server emits no `PUSH_DEVICE_STATE` for job-v1 devices.

Aggravator: the optimistic cell has `job_id: null`, so `clearAssignment` (which requires `current.job_id === jobId`) can never remove it. Server state stays correct and a page reload repaints, but the operator is shown a stale cell until then.

Suggestion: only record `renderedAssignments.set(...)` when `applyAssignment` returned true.

mdm-server/styly_mdm/static/push-jobs-v1.js
```diff
@@ -0,0 +1,510 @@
...
+    }).then(checkedJson).then(function (created) {
+      appendLog('Created job ' + created.job_id.slice(0, 8) + '; uploading artifact…', 'info');
+      if (created.state !== 'created' && created.state !== 'uploading') return created;
+      return nativeFetch(created.upload_url, { method: 'POST', body: staged.formData })
+        .then(checkedJson);
+    }).then(function (ready) {
+      if (!ready || ['ready', 'running', 'reconciling'].indexOf(ready.state) < 0) {
+        throw new Error('job is not dispatchable after upload');
+      }
+      nativeSend.call(socket, JSON.stringify({ type: 'PUSH_FILES', job_id: ready.job_id }));
```
hacha: **`PUSH_FILES` is dispatched on a socket captured before the upload.**

`socket` was captured before `POST /api/push-jobs` and before an artifact upload that runs for minutes on a large bundle. Lines 294 and 343 in this same file correctly use `currentAdminSocket` instead.

If the admin socket drops during the upload (index.html reconnects automatically), `send()` on a CLOSED socket throws `InvalidStateError`, and the catch at line 151 reports "Push/Sync job failed before dispatch" — while the job is fully created, uploaded and `ready` server-side. The operator is told the push failed and loses the on-screen trace of a job that actually just needs dispatching.

Suggestion: use `currentAdminSocket` here for consistency with lines 294/343.

mdm-server/styly_mdm/push_job_store.py
```diff
@@ -0,0 +1,1161 @@
...
+            raise StoreNotFound(job_id)
+        device_rows = conn.execute(
+            "SELECT * FROM push_job_devices WHERE job_id=? ORDER BY target_ordinal",
+            (job_id,),
+        ).fetchall()
+        aggregate = aggregate_device_states(row["state"] for row in device_rows)
+        devices: dict[str, Any] = {}
+        for row in device_rows:
+            fence = conn.execute(
+                "SELECT * FROM push_device_fences WHERE device_id=?", (row["device_id"],)
```
hacha: **A device-scoped fence is embedded into every job's device row.**

This query is scoped by `device_id` only, but its result is attached to each job's device entry. Once D1 is fenced by a new job, every job that ever targeted D1 — including long-completed `succeeded` jobs — carries that fence in its snapshot, and `push-jobs-v1.js:268` renders "fenced D1 (blocked by <other-job>)" plus a Reconcile button on those finished rows.

Cosmetic rather than a correctness defect, but it misattributes the block and re-decorates history that is already terminal.

mdm-server/styly_mdm/push_scheduler.py
```diff
@@ -0,0 +1,840 @@
...
+                        log.info(
+                            "Could not send recovery reconciliation for %s/%s",
+                            job_id,
+                            device_id,
+                        )
+                return
+            except asyncio.CancelledError:
+                raise
+            except StoreConflict:
+                continue
```
hacha: **`StoreConflict` is retried with no backoff and no attempt cap.**

The generic `except Exception` branch just below sleeps `_RUN_RETRY_DELAY`; this branch does not. Each iteration awaits the DB worker so the event loop is not blocked, but a persistently conflicting row becomes a hot loop against the single dedicated DB thread, starving every other job's transactions.

Suggestion: apply the same `_RUN_RETRY_DELAY` sleep, or bound the retry count.

mdm-client/app/src/main/java/com/styly/mdmclient/PushJobStore.kt
```diff
@@ -0,0 +1,68 @@
...
+        }
+    }
+
+    fun emptyState() = PushProtocol.State(null, emptyList(), emptyList())
+
+    private fun trim(state: PushProtocol.State): PushProtocol.State {
+        val cutoff = clock() - RECEIPT_RETENTION_MS
+        val retained = state.completedReceipts
+            .filter { it.result.completedAt <= 0L || it.result.completedAt >= cutoff }
+            .takeLast(MAX_RECEIPTS)
```
hacha: **Follow-up (non-blocking): `pendingResults` is unbounded.**

`trim()` bounds `completedReceipts` via `MAX_RECEIPTS` and the 7-day retention window, but `pendingResults` has no bound, and `PushJobCoordinator.replayPendingResults()` re-sends the entire list on every `REGISTERED`.

If ACKs are being lost, a frequently reconnecting device sends a linearly growing burst per reconnect and `state.json` grows without bound. Worth a bound here plus a cap on replay batch size, but it does not need to block this PR.

hacha (changes requested):
### Code review

Verified locally at `141cc74`: `mdm-server` `pytest` 355 passed, `mdm-client` `gradlew testDebugUnitTest` passed. All four safety properties in the description hold as written — SHA-256 and exact size are mandatory for job-v1 and checked before extraction, the 2 GiB bound is applied to both declared and actual uploaded bytes, transfer slots are released only by exact `(job_id, device_id, attempt)` identity, and legacy clients keep working (`MDM_ALLOW_LEGACY_PUSH` defaults to `1`). Transaction discipline is sound: `BEGIN IMMEDIATE` + WAL + `synchronous=FULL` on a single dedicated DB thread, with `_commit`/`_rollback` on every exit path.

This is unusually disciplined work for its size, and the design docs are excellent. Five inline comments below; the first is the only one I would treat as blocking.

Two notes outside the diff:

- `.github/workflows/` has no `on: pull_request` or `on: push` workflow, so `gh pr checks` reports nothing for this branch and neither test suite runs automatically. A +12,766-line change to the concurrency and persistence core is a strong argument for adding one.
- Coverage disclosure: I read `push_scheduler.py`'s dispatch/recovery path, `push_job_manager.py`'s `claim_next` and fence handling, `push_job_store.py`'s worker and transaction helpers, `push_runtime.py`'s `install()` and registration path, the console adapter, and the client's protocol/verification/persistence. The remainder — `push_jobs.py` validators, most of `push_runtime.py`'s message handling, and `PushJobCoordinator.kt` — I verified only by running the test suites.

0suu:
Thanks for the review. I addressed all six inline findings in the current HEAD
(1804c7e):

- Offline job-v1 fences can now be cleared safely on process replacement.
- Console assignment rendering no longer freezes on rejected assignments.
- Push dispatch uses the current admin WebSocket after upload.
- Device fences are shown only on the relevant/latest job.
- StoreConflict retries now include backoff.
- pendingResults is now bounded to 256 entries.

Could you please re-review this PR?
@hacha

hacha (approved):

PR #100, opened 2026-07-30
feat(client): allow manual server URL configuration

0suu:
## What changed

- Store the URL saved in Settings separately from the discovery cache.
- Treat manual configuration and Auto-Discovery as exclusive connection modes.
- In manual mode, keep the saved URL for the full connection window and retry only that URL. Discovery, the discovery cache, and the default URL are not used.
- In Auto-Discovery mode, discover on each attempt, cache successful results, and fall back to the discovery cache/default URL when no server replies.
- Add **Use Auto-Discovery** beside **Save & Connect**. It clears the manual override and previous discovery cache, then restarts the service for fresh discovery.
- Show only the manual override in the Settings field and validate entered `ws://` / `wss://` URLs.
- Snapshot the connection mode when each bounded connection window opens and update the operator/developer documentation.

## Why

The previous implementation stored manual input and discovery results in the same
`server_url` preference. Discovery could overwrite operator configuration, and the
Settings discovery flow could accidentally pin an automatically found server as a
manual endpoint.

Manual configuration must also remain authoritative while a venue server is still
booting. Retrying only once would abandon the operator's URL for the rest of the
connection window.

## Impact

- **Save & Connect** pins the entered URL and retries only that URL throughout each connection window.
- **Use Auto-Discovery** removes the pin and stale discovery cache, then immediately starts a fresh discovery-first window.
- Auto-Discovery caches successful results and uses the flavor-aware default URL only as the last fallback.
- Existing `connect_window_seconds` and `connect_retry_interval_seconds` limits are unchanged.
- Network transitions and established connection drops open a new window and re-evaluate the selected mode.

## Migration

Released clients did not record whether `server_url` came from Settings or discovery,
so that ambiguous legacy value is treated as a discovery cache. Operators who need it
pinned must re-enter it and tap **Save & Connect**. Tapping **Use Auto-Discovery** removes
that legacy cache before starting fresh discovery.

The unmerged draft Boolean key is intentionally ignored by using a fresh String key;
there is no production migration for an intermediate draft build.

## Validation

- All 61 Android ProdDebug unit tests passed.
- All 61 Android DevDebug unit tests passed.
- `app:assembleProdDebug` and `app:assembleDevDebug` passed.
- `git diff --check` passed.


hacha:
## Summary

Splitting the manual URL and the discovery cache into separate preference keys (`manual_server_url_value` / `server_url`) is the right fix, and the way it hooks into the state machine is sound: `WindowOpened` arms the one-shot exactly once per window, and a mid-window socket drop correctly does not re-arm it. The scheduler tests assert that invariant.

One blocking regression against `develop`, plus a few comment-level items.

---

## 🔴 Blocking: a manual URL is never retried after its first failure in a window

`WebSocketManager.kt` — `startAttempt()` / `getCachedServerUrl()`

**On `develop`:** `doConnect()` read `getServerUrl()` → the `server_url` key, which is exactly where `SettingsActivity.saveAndRestart()` wrote the operator's URL. So as long as discovery kept failing, **every** attempt cycle in the window connected to the manually configured URL.

**After this PR:** the manual URL is consumed once via `manualUrlAttempt.take()`, and every later cycle calls `doConnect(getCachedServerUrl(context))` → discovery cache, or else `DEFAULT_SERVER_URL` (the hardcoded `192.168.1.100`). **The operator's URL is never tried again for the rest of that window.**

The scenario this breaks is the one the connection-window feature exists for: the headset powers on before the server finishes booting. The first manual attempt fails, and the remainder of the window is spent against a fixed IP that is almost certainly wrong at the venue. A PR whose stated purpose is "discovery must not clobber operator config" ends up abandoning operator config after a single failed attempt.

**Suggestion:** make the fallback order `manual → discovery cache → default`. Extracting it as a pure `chooseFallbackUrl(manual, cache, default)` keeps it host-JVM testable and closes the coverage gap noted below.

## 🟡 The new timing budget is documented against the wrong fallback

The README / DEVELOPMENT.md note says a black-holed manual URL leaves "about 2 seconds for the cached fallback". With the defaults it is actually 3s manual + 2s retry + 3s discovery + 3s connect = 11s > the 10s window, and that tail is spent on `DEFAULT_SERVER_URL`, not on anything useful. Fixing the fallback order makes most of this new prose — and the advice to lengthen `connect_window_seconds` — unnecessary. Prefer fixing the ordering over documenting the caveat.

## 🟡 Test coverage is on the trivial part

The 40 new test lines cover a 19-line holder around a single nullable field. The PR's central invariant — **discovery writes `PREF_SERVER_URL` and never `PREF_MANUAL_SERVER_URL`** — and the fallback ordering have no coverage at all, and cannot get any here without Robolectric. Extracting the ordering decision as a pure function would put the tests where the risk is.

## 🔵 Comment-level

- **PR body overclaims.** The "flavor-aware complete URL placeholders" bullet is a net no-op: `dev/res/values/strings.xml` is byte-identical between `develop` and this branch, and the `main` hint is unchanged. Worth dropping, since release notes are derived from PR text.
- **The connected server is not visible in the UI.** The empty field is fine now that the label reads "Manual Server URL (optional)" with a hint, but when connected via discovery the status shows only `Connected`, so an operator cannot tell which server they are on. Including the URL in the connected status would help in the field.
- **No CI on this branch.** The "61 tests passed / assemble passed" claims were not independently re-run during this review (Gradle dependency resolution needs hosts outside the review sandbox's allowlist).

---

Verified separately: no dangling references remain to the removed `getServerUrl` / `setServerUrl` / `discover_button` / `discover_server`, and the `Request.Builder().url()` validation is correct for `ws://` / `wss://` (OkHttp normalizes them to http/https).

hacha:
LGTM

PR #101, opened 2026-07-30
feat: uninstall an app on managed devices from the console (#63)

### LINKED ISSUE #63: Removing an app installed via STYLY-MDM requires physical device access
## Problem

STYLY-MDM can push and remotely install content APKs to a fleet (`INSTALL_APK`/`EXECUTE_INSTALL`), but there is no way to remove an installed app afterward. Today the only path is per-device physical access (`adb uninstall <package_name>`) — at exactly the fleet size where MDM was adopted to avoid per-device handling.

## Current state

- `INSTALL_APK`/`EXECUTE_INSTALL` installs by APK file + `package_name` with hash verification.
- `LAUNCH_APP`/`EXECUTE_LAUNCH` (`mdm-server/styly_mdm/server.py:1880-1922`) launches an already-installed app by `package_name` — the closest structural template for a delete command (validate `package_name`, resolve `target_ids`, fan out, reply with sent/target counts).
- `RETIRE_DEVICE`/`EXECUTE_SELF_UNINSTALL` (#49) already uninstalls a package via ToBService, but only the MDM client's own package — not arbitrary target apps.
- The client already calls the silent-uninstall API in `mdm-client/app/src/main/java/com/styly/mdmclient/MdmClientService.kt:719,755` (`pbsControlAPPManger(PACKAGE_SILENCE_UNINSTALL, packageName, ...)`), but it is hardcoded to the guard/self package names only.
- No per-device installed-app inventory exists anywhere, so the console has no list to pick a target app from — the operator would need to know the package name already.
- App identity is consistently `package_name` (string) throughout the codebase (`LAUNCH_APP`, `SET_STARTUP_APP`, `VERIFY_APK`), not a numeric app ID.

## Approaches

- **New `DELETE_APP`/`EXECUTE_UNINSTALL` command (preferred).** Follow the `LAUNCH_APP` structural pattern: admin supplies `package_name` + target device(s), server fans out, client calls `pbsControlAPPManger(PACKAGE_SILENCE_UNINSTALL, package_name, ...)` for the given package (generalizing the retire-only call), and replies with a per-device result.
- **Status quo / fallback.** Manual `adb uninstall` per device.

## Open questions

1. Per project convention for PICO Enterprise SDK behavior, confirm via Context7 (`/websites/business_picoxr_us_doc`) and/or a device spike that `PACKAGE_SILENCE_UNINSTALL` works for arbitrary third-party packages, not just the guard/self package used by retire (#49).
2. Does the console need a confirmation step, given the operation is immediate and remote?
3. Should this record/require a package name manually, or does it depend on a future installed-app inventory feature?

hacha:
Closes #63.

## Problem

Content APKs could be pushed and silently installed to a fleet, but never removed. The only path was physical access per headset and `adb uninstall <package_name>` — exactly the per-device handling MDM was adopted to avoid.

## What this adds

`DELETE_APP` → `EXECUTE_UNINSTALL` → `DELETE_APP_RESULT` (ack: `DELETE_APP_SENT`). The fan-out is the `LAUNCH_APP` shape — online targets only, un-throttled, since nothing is transferred — and on the device it is the silent-package API the install and retire paths already use: `pbsControlAPPManger(PACKAGE_SILENCE_UNINSTALL, package_name, …)`.

The **result** follows `INSTALL_RESULT`, not retire. Retire's "silence is success" machinery exists only because a self-uninstall kills the process before the `IIntCallback` can matter; uninstalling somebody else's package leaves the client alive, so the callback always fires and the result is reported directly.

## Two rules that carry real consequences

**STYLY-MDM's own packages are refused**, on the server (`PROTECTED_PACKAGES`) *and* again on the client against its own package names, so a drift between the two lists degrades to a device-side refusal. Removing either here would skip the entire ordered retire ceremony — guard first, keep-alive deregistration, the `SELF_UNINSTALL_STARTING` announcement, recovery on failure — and leave a device unmanaged with no console-side way back. `RETIRE_DEVICE` stays the only sanctioned path.

**Startup App interaction.** A target that is the device's startup app has that setting cleared *before* the uninstall and restored if it fails, so a failed operation changes nothing. The server honours `startup_app_cleared` only on `status: success`, and does not depend on it alone: a successful uninstall also clears its record whenever the record still names the removed package — `SET_STARTUP_APP` is recorded server-side at dispatch, before the device confirms, so the two sides can drift.

## Notes

- The operator types the package name. An installed-app inventory to pick from is a separate follow-up, not a dependency (issue #63 open question 3).
- The console gates the action behind a confirmation checkbox plus a `confirm()` naming the package and the target devices, mirroring Retire and Power Control.
- Client-side hardening: the pre-flight PackageManager call runs on the worker thread (not the WebSocket reader), the callback wait is bounded, and leaving the in-flight set is the once-only report guard so a late callback cannot produce a second result.
- **This needs a MINOR bump** — server and client must move together.

## Verification

- `PYTHONPATH=. python -m pytest tests/ -q` → **234 passed** (222 before, +12 new in `tests/test_delete_app.py`).
- `./gradlew :app:compileProdDebugKotlin --rerun-tasks` → BUILD SUCCESSFUL.
- Console inline JS parsed with `node --check`; no Japanese characters in `index.html`.
- `message-flow.drawio.svg`: both the rendered SVG and the embedded mxfile XML parse; `viewBox` grown and lifelines extended for the new phase.

### Still device-only

The issue's open question 1 is resolved from the codebase rather than Context7 — the retire path at `MdmClientService.kt:816-826` already removes `GuardLink.GUARD_PACKAGE` (a package other than the caller) via `PACKAGE_SILENCE_UNINSTALL` with a real `IIntCallback`, and that path is device-verified. Context7 (`/websites/business_picoxr_us_doc`) only indexes the Enterprise API intro page and has no `pbsControlAPPManger` reference.

Residual: the guard is a companion app we ship, so **"works on an arbitrary third-party content APK" still wants one on-device spike**, along with the not-installed path and the Startup App column clearing. The protected-package refusal is already covered by tests and needs no device.


hacha:
Review result: blocking findings, not merge-ready yet.

1. **P1: late uninstall callbacks can corrupt or mask a later operation.** In `mdm-client/app/src/main/java/com/styly/mdmclient/MdmClientService.kt:506`, the timeout path restores startup-app prefs and reports failure, and `finishUninstall()` releases the only in-flight key at `MdmClientService.kt:535`. The original SDK callback object can still fire later. If it reports success after the timeout, the result is dropped while the startup app may already have been restored to a now-uninstalled package. If an operator retries the same package after the timeout, that late callback can consume the retry key and send the old result as if it belonged to the newer command. This needs a per-operation token/correlation guard, or equivalent scoping so timeout and callback cannot race through the same package-only key.

2. **P1: startup-app clearing/restoring is not scoped to the uninstall operation.** On the server, `server.py:1486-1496` clears the current startup-app record whenever a successful `DELETE_APP_RESULT` has `startup_app_cleared`, even if the current record no longer names the result package. `handle_set_startup_app()` records a new startup app immediately at `server.py:2895-2906`, so this sequence loses a valid newer setting: uninstall old startup app -> set a new startup app before the uninstall result returns -> old result arrives with `startup_app_cleared` -> server clears the new record. The client has the matching reverse problem at `MdmClientService.kt:469-508`: it snapshots the old setting and restores it on failure/timeout, even if a later `SET_STARTUP_APP` already changed prefs. The clear/restore should only apply when the current startup-app package still matches the uninstall operation it belongs to.

3. **P2: duplicate same-package commands produce no terminal result.** `MdmClientService.kt:441-443` ignores a second `EXECUTE_UNINSTALL` for the same package while one is in flight and sends no `DELETE_APP_RESULT`. The server still acks each admin dispatch with `DELETE_APP_SENT` at `server.py:2078-2083`, so the operator can see “sent” for a command that never receives any per-device completion or failure. Please send a deterministic terminal result such as `fail/already in flight`, or make the server/client protocol explicitly correlate and coalesce duplicates.

Verification I ran locally:
- `PYTHONPATH=mdm-server python -m pytest -q mdm-server/tests/test_delete_app.py` -> 12 passed
- `PYTHONPATH=mdm-server python -m pytest -q` -> 234 passed
- inline console JS parse via Node `new Function(...)` -> OK
- Japanese-character guard on `mdm-server/styly_mdm/static/index.html` -> no hits
- `git diff --check origin/develop...HEAD` -> OK

Notes: `gh pr checks 101` reported no checks on this branch. Android compile was not rerun in my local environment because no Java Runtime is installed. Context7 for `/websites/business_picoxr_us_doc` confirms PICO Enterprise API includes high-risk silent app install/uninstall and the `pico_advance_interface` requirement, but it did not expose detailed `pbsControlAPPManger` docs.

hacha:
All three findings addressed in 29f933d. Thanks — #1 and #2 were both real races in the window between dispatch and result, and neither was reachable from the tests I had.

**1. Late callbacks (P1).** The in-flight guard was keyed on the package alone, so once the 60 s wait timed out and released it, the original SDK callback could still fire and either drop its result after the startup app had been restored to a now-gone package, or consume a retry's claim and report the old outcome as that retry's. Each attempt now carries a token (`uninstallsInFlight: package -> op`); the callback checks ownership before touching any state, and releasing the claim happens inside the same guarded step as reporting, so a stale attempt can neither report twice nor free a newer one's claim.

While fixing this I noticed the timeout path was also *guessing* — it reported failure and restored the startup app without checking what had actually happened. The package list is the ground truth at that point, so it is now read: gone means the uninstall did happen, and it is reported as a success.

**2. Startup-app scoping (P1).** Correct on both ends, and your sequence is exactly the one that broke. Server: `startup_app_cleared` is now informational; the record is cleared only when it still names the removed package, so a `SET_STARTUP_APP` issued mid-uninstall survives a late result. Client: the restore now happens only when nothing else has set a startup app since, so a failed uninstall cannot overwrite a newer setting with the package it just failed to remove.

**3. Duplicate command (P2).** Now answered with a terminal `DELETE_APP_RESULT status=fail` ("An uninstall of this package is already in flight"), rather than silence against a `DELETE_APP_SENT` the operator already saw.

Docs updated to match (the `DELETE_APP_RESULT` rows, the *Remote App Uninstall* section, and the diagram label).

Verification:
- `PYTHONPATH=mdm-server python -m pytest -q mdm-server/tests` -> **235 passed** (+1: a new e2e case pinning that a startup app set mid-uninstall survives a late result carrying `startup_app_cleared`).
- `./gradlew :app:compileProdDebugKotlin --rerun-tasks` -> BUILD SUCCESSFUL.
- Both representations of `message-flow.drawio.svg` still parse.

The device-only items are unchanged: uninstalling an arbitrary third-party APK, the not-installed path, and the Startup App column clearing.

PR #104, opened 2026-08-03
fix(client): preserve PICO USB power-off setting

### LINKED ISSUE #103: Powering off the PICO device enables the USB disconnect power-off setting
### Description

When the PICO device is powered off, the "Power off when USB is disconnected" setting in the device's system settings is automatically enabled.

The issue reproduces whether or not the PICO device is connected to a power source.

Rebooting does not reproduce the issue.

0suu:
## Summary

- Stop enabling PICO's persistent "Power off when USB is disconnected" setting during remote Power Off.
- Keep the existing `DEVICE_CONTROL_SHUTDOWN` call.
- Update the UI warning and developer documentation to reflect that power-off while charging depends on the device's existing PICO setting.

## Root cause

The client unconditionally called `pbsControlSetPowerOffwithUSBCable(S_ON, 0)` before every remote shutdown. The setting is device-wide and persistent, so it remained enabled after power-off even when the headset was not connected to USB.

Fixes #103

## Validation

- `./gradlew :app:compileDevDebugKotlin`
- `./gradlew :app:testDevDebugUnitTest`
- `git diff --check`

Both Gradle checks passed. Physical PICO verification is still required, especially for shutdown while USB-connected.

### DIFF
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -926,17 +926,12 @@ gated by the `pico_advance_interface` manifest flag the client already declares.
 permission and no SDK addition were needed; shipping the two command handlers is what
 requires a client update + redeploy.
 
-**Power off while charging.** A PICO headset refuses a full power-off while a USB/charging
-cable is connected unless the device-wide *"power off with USB cable"* setting is enabled —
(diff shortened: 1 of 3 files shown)


0suu:
Verification is required.

hacha:
LGTM

PR #107, opened 2026-09-14
Use Device-ID-Provider GUIDs for device identity

### LINKED ISSUE #65: Matching an MDM-managed headset to its STYLY records requires hand-mapping serial numbers to Device IDs
## Problem

STYLY-MDM identifies every device by its hardware serial number: `mdm-client` reads the PICO ToBService `EQUIPMENT_SN` (falling back to `Build.SERIAL`) and sends it as `device_id` in `REGISTER` (`mdm-client/app/src/main/java/com/styly/mdmclient/WebSocketManager.kt:317-325`, used at `:247` and `:294`), and the server keys `device_registry` by that value.

The rest of the STYLY ecosystem identifies the same physical headset by a stable, pseudo-anonymous GUID produced by [Device-ID-Provider](https://github.com/styly-dev/Device-ID-Provider). Nothing on the MDM side ever sees that GUID, so:

- **STYLY-side records can't be joined.** Correlating a headset in the console with its STYLY platform records (content delivery, licensing, accounts) means a human maintaining a serial -> GUID mapping by hand.
- **Analytics and logs can't be correlated.** Telemetry keyed by the STYLY Device ID can't be joined against the operational data MDM already holds (battery, client version, `last_seen`).
- **Device identity is not standardized.** MDM is the only STYLY component keyed by serial. The `Build.SERIAL` fallback also returns `"unknown"` for non-privileged apps on Android 10+ (API 29+), so on any non-PICO hardware the fallback yields a non-unique id.

## What we need

`mdm-client` should report the STYLY Device ID for each device, and the server/console should store and surface it alongside the serial.

The client must also be able to **mint** the ID: a headset fresh from kitting can reach MDM before any STYLY app has ever run, so MDM can legitimately be the first writer. Read-only access is not sufficient.

## Where the ID actually lives (relevant to every approach below)

The ID is not stored inside the library. On Android it is a 1x1 PNG in the public MediaStore under `Pictures/Device-ID-Provider/`, where **the GUID is the filename** (`<guid>.png`), and the oldest entry by `date_added` wins. The library is only an accessor: any app reading the same MediaStore location gets the same ID.

That makes *reading* straightforward. **Minting** is the hard part — create the PNG, insert it, re-query to converge on the oldest under races — and it is exactly where two independent implementations can silently disagree and mint *different* IDs for one device.

### 1. Extract the Android logic upstream into a real Android library

Have Device-ID-Provider expose its Android implementation as an `.aar` / androidlib with actual Java/Kotlin source, consumed by both the Unity package and `mdm-client`.

Today `Packages/com.styly.device-id-provider/Plugins/Android/DeviceIDProvider.androidlib` contains only an `AndroidManifest.xml` (permission declarations); all logic lives in C# (`Runtime/Providers/AndroidDeviceIdProvider.cs`) and reaches Android through `AndroidJavaObject`.

- Pros: one canonical implementation of the mint path; no Unity runtime in `mdm-client`; benefits any future native STYLY app.
- Cons: cross-repo change in Device-ID-Provider plus release coordination; blocks this issue on upstream.

### 2. Reimplement the scheme natively in Kotlin inside mdm-client

- Pros: self-contained, no Unity runtime, works directly from the existing background service.
- Cons: a second implementation of the mint / race-convergence path — divergence would be silent and would hand one device two identities; must be kept in sync with upstream changes.

### 3. Embed Device-ID-Provider via Unity as a Library

- Pros: reuses the canonical implementation verbatim, no reimplementation.
- Cons:
  - `Runtime/Internal/AndroidBridge.cs` resolves its context via `com.unity3d.player.UnityPlayer.currentActivity`. `mdm-client` registers from a headless, boot-launched foreground service (`MdmClientService`, `foregroundServiceType="connectedDevice"`) with no Activity on screen, so obtaining the ID would require starting a Unity Activity at boot — conflicting with the Startup App auto-launch behaviour operators rely on.
  - Adds the Unity runtime (tens of MB) and a Unity 6000.0+ toolchain to a small Kotlin client, to obtain a single GUID.

Suggested direction: **1** if upstream can take it, otherwise **2**.

## Open questions

- **Permission on a headless client.** The package declares `READ_EXTERNAL_STORAGE` (`maxSdkVersion=32`) / `READ_MEDIA_IMAGES`, and the C# path blocks on an interactive permission prompt — unworkable with no UI. `mdm-client` is not a Device Owner (no `DevicePolicyManager` usage), so it cannot self-grant, and the PICO Enterprise docs do not document a silent runtime-permission grant. It does already hold `MANAGE_EXTERNAL_STORAGE`, which may make `Pictures/Device-ID-Provider/` directly accessible — needs verifying on a real PICO device.
- **Serial vs GUID as the key.** Suggest keeping the serial as the registry primary key and carrying the Device ID as an additional field, at least initially, since the GUID can fail to resolve.
- **Failure behaviour.** Registration must not break when the ID is unavailable (permission denied, API < 29).

0suu:
## Summary
Switch device identification from hardware serial numbers to GUIDs provided by Device-ID-Provider.

## Changes
- Resolve device IDs asynchronously and retry after permissions are granted.
- Show unresolved connections in “Need attention,” with reboot and power-off controls for supported clients.
- Add shortened device ID display and full-ID copying.
- Improve registration acknowledgement handling and command eligibility checks.

## Upgrade notes
- **Automatic migration is not included.** Labels, groups, and Push history are not transferred from legacy serial IDs to GUIDs.
- **Register each device’s GUID and physical name before upgrading.**
- Complete pending Push/Sync jobs targeting legacy serial IDs, then upgrade the server before updating clients.
- Updated clients register as new devices under their GUIDs. Legacy clients are limited to APK installation and update commands.
- Unlike the original issue proposal, hardware serial numbers are not used as a fallback. Clients that cannot obtain a GUID appear in “Need attention” until the ID is resolved.

Closes #65


hacha:
## Review

Reviewed the diff (device identity migration to Device-ID-Provider GUIDs). Overall the approach looks sound, but found a few issues worth addressing before merge, plus some maintainability notes.

### High priority — functional bugs

**1. `DeviceIdentityResolver.kt:106` — no retry path for timeout/IO errors**
`retryAfterPermissionGranted()` only retries when the last resolution failed with `ACCESS_DENIED`. If `getOrCreateAsync` times out (`IO_ERROR`) — e.g. Device-ID-Provider is slow to start on cold boot — the resolver stays stuck in `Unavailable` for the process lifetime. There's no caller path that retries on timeout, so install/push/telemetry stay blocked until the app is restarted.

**2. `MdmClientService.kt:629` — reboot/power-off can silently no-op on socket reconnect**
After sending the `*_RESULT: accepted` ack for `EXECUTE_REBOOT`/`EXECUTE_POWER_OFF`, the code waits up to 2s for outbound flush via `CommandContext.isActive()`. If the socket reconnects during that window (e.g. brief Wi-Fi blip unrelated to the reboot itself), `isActive()` compares against the old socket and returns `false`, so `awaitOutboundFlushForCommand` returns `false` and the function returns *before ever calling* `pbsControlSetDeviceAction`. No failure frame is sent, so the admin console shows the command as delivered while the device never actually reboots/powers off.

### Medium priority — data inconsistency

**3. `server.py:1308` vs `push_jobs.py:333` — two different parsers for the same `capabilities` field**
`_bounded_capabilities()` (used to populate `devices[device_id]["capabilities"]` for the console) skips individual malformed entries, while `push_jobs.parse_capabilities()` (used to build `session.capabilities` for push routing) is all-or-nothing — one malformed entry drops the whole list. A REGISTER frame with e.g. `["push_job_id_v1", ""]` ends up showing the device as v1-push-capable in the console while actually being routed onto the legacy push path. Worth unifying on one parsing rule.

### Low priority — maintainability / duplication

**4. `push_runtime.py:474`** — legacy-identity devices are blocked from push jobs via an inline `identity_kind == "legacy"` check here, duplicating what `device_policy.command_allowed()` already enforces on the WebSocket dispatch path. Risk of the two gates drifting apart if the allow-list changes later.

**5. `push_runtime.py:161`** — privileged-command authorization in `RuntimeWebSocketResponse.send_str` is done by string-matching the serialized `type` (prefix `EXECUTE_` + a hardcoded `{"SET_STARTUP_APP", "CLEAR_STARTUP_APP", "PUSH_RECONCILE_REQUEST"}` set). A future privileged command that doesn't match this pattern would silently bypass the check.

**6. `server.py:173`** — `PROVISIONAL_RESULT_COMMANDS` hardcodes the `REBOOT_RESULT`/`POWER_OFF_RESULT` → `EXECUTE_*` mapping instead of deriving it from `device_policy.PROVISIONAL_POWER_COMMANDS`, which is already imported here. A future third provisional power command would need this dict updated by hand in a second place.

---

hacha (changes requested):
Requesting changes based on the review above — items #1 and #2 are functional bugs that should be fixed before merge:

- `DeviceIdentityResolver.kt:106` — no retry path when identity resolution fails with `IO_ERROR`/timeout, leaving the resolver permanently stuck.
- `MdmClientService.kt:629` — reboot/power-off can silently no-op if the socket reconnects during the post-ack flush wait, with no failure reported to the console.

Item #3 (capabilities parsing divergence between `server.py` and `push_jobs.py`) is also worth fixing in this PR since it affects push routing correctness. Items #4–#6 are lower-priority maintainability notes and can be follow-ups.

0suu:
@hacha 
Thank you for the review. Here are my responses to each point.

1. Device ID resolution checks readiness every 250 ms for up to 30 seconds, then stops if resolution still fails. This is intentional, as delayed automatic recovery could trigger processing at an unintended time. Even after a timeout, an administrator can trigger a de
2. Canceling power operations when the connection becomes invalid is intentional. The UI also states that “accepted” does not mean execution has completed.
3. Fixed. Registration and Push routing now share the same capability parser.
4. The checks at job creation and immediately before dispatch serve different purposes, so I think keeping both is reasonable.
5. Fixed. Instead of checking command names, all device-bound messages except registration responses and ERROR now undergo connection ownership, registration state, and command permission checks.
6. The explicit mapping remains readable with just two entries, so I think keeping it as-is is reasonable.

